PR #14, opened 2022-09-10
error handling, global session

AlirezaMorsali:


invertedai_drive/drive.py
```diff
@@ -1,10 +1,12 @@
-from requests.api import request
 from invertedai_drive.utils import Client
+from invertedai_drive.error import TryAgain
```
Ruishenl, line 2: My first instinct is to call it 'exceptions' instead of 'error'.

invertedai_drive/drive.py
```diff
@@ -29,7 +31,8 @@ class Drive:
     def __init__(self, config) -> None:
         self.location = config.location
         self.config = config
-        self.client = Client(self.config.api_key)
+        # self.client = Client(self.config.api_key)
```
Ruishenl, line 34: Remove commented line

invertedai_drive/utils.py
```diff
@@ -167,3 +258,42 @@ def _make_box_layout(self):
...
+                val = val.decode("utf-8")
+            # Check if val is already a string to avoid re-encoding into ascii.
+            if not isinstance(val, str):
+                val = str(val)
+            if re.search(r"\s", val):
+                val = repr(val)
+            # key should already be a string
+            if re.search(r"\s", key):
+                key = repr(key)
+            return "{key}={val}".format(key=key, val=val)
```
Ruishenl, line 297: switch to f string

invertedai_drive/utils.py
```diff
@@ -167,3 +258,42 @@ def _make_box_layout(self):
...
+        log_file: bool = False,
+    ) -> None:
+
+        log_level = level if type(level := logging.getLevelName(level)) == int else 40
+        super().__init__(name, log_level)
+        if consoel:
+            consoel_handler = logging.StreamHandler()
+            self.addHandler(consoel_handler)
+        if log_file:
+            file_handler = logging.FileHandler("iai-loggs.txt")
```
Ruishenl, line 278: typo loggs ? and also consider rename it to `iai.log` instead

invertedai_drive/utils.py
```diff
@@ -167,3 +258,42 @@ def _make_box_layout(self):
             margin="0px 10px 10px 0px",
             padding="5px 5px 5px 5px",
         )
+
+
+class IAI_Logger(logging.Logger):
```
Ruishenl, line 263: Rename to IAILogger

invertedai_drive/utils.py
```diff
@@ -167,3 +258,42 @@ def _make_box_layout(self):
...
+    def __init__(
+        self,
+        name: str = "IAI_Logger",
```
Ruishenl, line 266: also here

invertedai_drive/utils.py
```diff
@@ -167,3 +258,42 @@ def _make_box_layout(self):
...
+        level: str = "WARNING",
+        consoel: bool = True,
+        log_file: bool = False,
+    ) -> None:
+
+        log_level = level if type(level := logging.getLevelName(level)) == int else 40
```
Ruishenl, line 272: Since default level is WARNING I assume it makes more sense to fallback to 30 ? unless you think it should all default to ERROR

invertedai_drive/drive.py
```diff
@@ -160,6 +165,6 @@ def _validate_and_tolist(input_data: dict, input_name: str):
                 return self.client.run(model_inputs)
             except Exception as e:
                 # TODO: Add logger
-                print("Retrying")
+                iai.logger.info("Retrying")
```
Ruishenl, line 168: i think warning is better suited here

invertedai_drive/utils.py
```diff
@@ -81,24 +85,111 @@ def _request(
...
+        except (KeyError, TypeError):
+            raise error.APIError(
+                "Invalid response object from API: %r (HTTP response code "
+                "was %d)" % (rbody, rcode),
+                rbody,
+                rcode,
+                resp,
+            )
+
+        if "internal_message" in error_data:
```
Ruishenl, line 124: what is internal_message here ?

invertedai_drive/drive.py
```diff
@@ -63,11 +66,13 @@ def initialize(
...
-            except Exception as e:
-                # TODO: Add logger
-                print("Retrying")
+            except TryAgain as e:
                 if timeout is not None and time.time() > start + timeout:
                     raise e
+                iai.logger.info(
+                    iai.logger.logfmt("Waiting for model to warm up", error=e)
+                )
+                breakpoint()
```
Ruishenl, line 75: remove breakpoint

Ruishenl (changes requested):

invertedai_drive/drive.py
Ruishenl: Just keep it it’s fine. Not important

invertedai_drive/utils.py
AlirezaMorsali: This would be a place for us to send error data from the server, in case there is some discrepancy in the client's data.

Ruishenl (approved):

PR #15, opened 2022-09-15
renaming drive to invertedai

AlirezaMorsali:


pyproject.toml
```diff
@@ -1,12 +1,12 @@
 [tool.poetry]
 name = "invertedai_drive"
```
Ruishenl, line 2: 'invertedai' should be the new package name

Ruishenl (approved):

PR #36, opened 2022-10-17
Stateful client wrapper

adscib:
I sketched an implementation of a client-side stateful wrapper to simplify co-simulation. The intention is to support the most straightforward use case only, letting the users get more fancy by either inheriting from this class or calling the API directly. I'll wait until the upcoming API changes are done before trying to use this for demos.


invertedai/simulation.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    def __init__(
+            self,
+            # all arguments to initialize are also given to this constructor
+            location: str = "CARLA:Town03:Roundabout",
+            agent_count: int = 1,
+            # some further configuration options
+            monitor_infractions: bool = False,
+            render_birdview: bool = False,
+            ego_agent_mask: Optional[List[bool]] = None,
```
Ruishenl, line 19: Instead of using mask with for loops, we can  just use ego_agents_indices and do something like map(self._agent_states.__getitem__, self._ego_agents_indices) to mutate the agent_states directly

adscib: I guess, but what would be the advantage?

adscib: We're limited to 100 agents anyway, so I don't think we need to worry about that. I'd rather do what's easier to read.

Ruishenl (approved):

AlirezaMorsali (approved):

Issue #30, opened 2022-10-07
Mock API version


adscib:
We should provide a mock interface that people can use to test the integration without having to contact our servers and incur charges. There are various ways in which this could be implemented, but I generally think the mock interface should share as much of the code as possible with the real interface to avoid them getting out of sync. It may be entirely reasonable to control this through an environment variable.

As for what the mock should do, it should be simple but still have noticeable effect. For DRIVE we could just keep constant velocity and update positions accordingly (and perhaps add a random bit of acceleration so that it has some stochastic behavior). For INITIALIZE it's a little trickier, maybe we could place the agents in some random locations around the center of the map. Unfortunately they won't necessarily be valid in CARLA, but I don't see a simple way around it.

PR #37, opened 2022-10-17
Mock API

adscib:
As described in #30. I opted to use an environment variable to control whether the mock API is used and for now I'm just returning zeros or current state deterministically everywhere. I also updated the Demo-Drive notebook to use the mock API when no key is specified.


examples/Demo_Drive.py
```diff
@@ -21,7 +21,12 @@
 parser.add_argument("--location", type=str, default="iai:ubc_roundabout")
 args = parser.parse_args()
 
-iai.add_apikey("")
+api_key = args.api_key or os.environ.get('IAI_API_KEY', None)
```
Ruishenl, line 24: .get() will default to None

invertedai/api_resources.py
```diff
@@ -30,6 +34,10 @@
 
 TIMEOUT = 10
 
+mock_api = bool(os.environ.get("IAI_MOCK_API", False))
```
Ruishenl, line 37: small suggestions.. strtobool is available to support slightly more cases
```
from distutils.util import strtobool
strtobool(os.environ.get("IAI_MOCK_API", "false")
```

True values are y, yes, t, true, on and 1; false values are n, no, f, false, off and 0

examples/Demo_Drive.py
adscib: I like to be explicit anyway.

invertedai/api_resources.py
adscib: done

invertedai/utils.py
```diff
@@ -33,6 +33,11 @@ def add_apikey(self, api_token: str = ""):
             raise error.InvalidAPIKeyError("Empty API key received.")
         self.session.auth = APITokenAuth(api_token)
 
+    def use_mock_api(self, use_mock: bool = True) -> None:
+        api_resources.mock_api = use_mock
+        if use_mock:
+            print('Using mock Inverted AI API - predictions will be trivial')  # TODO: replace with a suitable logger
```
adscib, line 39: Do we have some uniform mechanism in place to issue warnings like this?

Ruishenl: iai.logger

adscib: Is that safe in terms of circular imports?

Ruishenl: i think all the current reference of iai.logger happens inside the function so i never ran into any issue. We are already using it in this module

adscib: OK. Did I use it correctly or should I do some extra formatting?

Ruishenl: looks good to me

Ruishenl (approved):

Ruishenl:
Feel free to merge once @AlirezaMorsali  is okay with it

AlirezaMorsali:
I can fix the conflicts if you are working on something else.

adscib:
Go ahead - I'm working on the docs.

PR #42, opened 2022-10-20
Python library tweaks

adscib:
I have moved a lot of stuff around for a cleaner structure, adjusted some argument names and default values, and converted type aliases into their own classes, mostly to give their documentation a place to live. I checked that a basic demo script works with the mock api, the other scripts will likely need minor adjustments. Finally, I added lots of docstrings everywhere.

Nothing done here requires modifying the REST API.


AlirezaMorsali:
> 

update-interface

invertedai/common.py
```diff
@@ -0,0 +1,177 @@
...
+        Decode the image and save it to the specified path.
+        Requires numpy and cv2 to be available, otherwise raises `ImportError`.
+        """
+        image = self.decode()
+        import cv2
+        cv2.imwrite(path, image)
+
+
+
+class TrafficLightState(Enum):
```
Ruishenl, line 90: needs to be (str, Enum) to be serializable

Ruishenl:
```
class ControlStateType(StrEnum):
    none = auto()
    green = auto()
    yellow = auto()
    red = auto()

```

PR #51, opened 2022-10-21
Carlasim

AlirezaMorsali:


adscib (changes requested):
I only reviewed changes, will start going through the whole example as well.

simulators/carla_simulator.py
```diff
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
```
adscib, line 1: I think everything CARLA-related should go into one folder, such as `examples/carla`.

simulators/carla_simulator.py
```diff
@@ -26,7 +25,9 @@ class RecurrentState:
     It should not be modified, but rather passed along as received.
```
adscib, line 25: Why are those dataclasses redefined instead of being imported from `invertedai`?

Carla_examples/Carla-Scenario.py
```diff
@@ -42,15 +35,15 @@
     max_cars_in_map=args.max_cars_in_map,
 )
 
-iai.add_apikey("")
+iai.add_apikey(args.api_key)
```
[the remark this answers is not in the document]
AlirezaMorsali: When `invertedai` is imported, the API key is set from the "IAI_API_KEY" if it exists. This is for ease of use if someone is actually doing development. 
`iai.add_apikey("xxxx")` is for sake of demo mostly, I beleive.

PR #52, opened 2022-10-21
Document env var

Ruishenl:
Screenshot from 2022-10-21 13-31-37


adscib (approved):

adscib:
The user guide refers to the mock API somewhere, so maybe we can link to this new page there.

PR #53, opened 2022-10-21
Improve error handling

Ruishenl:


adscib (approved):
Cool. I don't have any list for what error conditions we should cover, so we'll just go with what you did.

PR #54, opened 2022-10-21
Add local simulator mockup to minimal example

adscib:
I got things working with the mock API, but wasn't able to run a local server to test it yet. This is now a complete example that could potentially go on the website. When we have birdviews in initialize, we could also store a gif here, since currently it doesn't show anything.


adscib:
Note that this includes a tiny change to the library, where I made `npc_states` a property to be consistent with `ego_states`.

PR #63, opened 2022-10-24
add birdview and infractions to initialize + add random_seed to co-simulator

AlirezaMorsali:
Adding bird view image, infractions, and is_inside_supported_area to initialize


invertedai/cosimulation.py
```diff
@@ -33,30 +33,45 @@ def __init__(
...
+            get_birdview=get_birdview,
+            get_infractions=monitor_infractions,
+            **kwargs,
+        )
+        if monitor_infractions and (response.infractions is not None):
+            # TODO: package infractions in dataclass
+            self._infractions = (
+                [inf.collisions for inf in response.infractions],
+                [inf.offroad for inf in response.infractions],
+                [inf.wrong_way for inf in response.infractions],
```
adscib, line 52: Does this still make sense? I think we have the required dataclass now, which we're unnecessarily deconstructing here.

invertedai/cosimulation.py
```diff
@@ -33,30 +33,45 @@ def __init__(
...
+            self._birdview = response.birdview
         if ego_agent_mask is None:
             self._ego_agent_mask = [False] * self._agent_count
         else:
             self._ego_agent_mask = ego_agent_mask[: self._agent_count]
         # initialize might not return the exact number of agents requested,
         # in which case we need to adjust the ego agent mask
         if len(self._ego_agent_mask) > self._agent_count:
-            self._ego_agent_mask = self._ego_agent_mask[:self._agent_count]
+            self._ego_agent_mask = self._ego_agent_mask[: self._agent_count]
```
adscib, line 74: Why the extra space?

invertedai/api/initialize.py
```diff
@@ -93,27 +121,40 @@ def initialize(
         if agent_attributes is None
         else [state.tolist() for state in agent_attributes],
         traffic_light_state_history=traffic_light_state_history,
+        get_birdview=get_birdview,
+        get_infractions=get_infractions,
```
adscib, line 125: Don't you also need to adjust the mock api output to include those?

adscib (changes requested):
Are you trying to fit the code lines into 80 characters or something? Not that this is necessarily wrong, but we should be consistent about formatting.

invertedai/cosimulation.py
AlirezaMorsali: I had the black code formatter on.
To be consistent with you and Ruishen, I will use pep8 from now on and I guess you have max-line-length=119 and not 79.

invertedai/cosimulation.py
```diff
@@ -100,17 +115,32 @@ def ego_states(self):
...
-    def infractions(self) -> List[InfractionIndicators]:
+    def infractions(self) -> Optional[List[InfractionIndicators]]:
         """
         If `monitor_infractions` was set in the constructor,
         lists infractions currently committed by each agent, including ego agents.
         """
         return self._infractions
 
+    @property
+    def infraction_rates(self) -> Optional[Tuple[float, float, float]]:
```
adscib, line 126: I'm not too keen on having this here. It doesn't seem particularly useful to me to average collisions across agents.

invertedai/cosimulation.py
```diff
@@ -23,36 +24,50 @@ class BasicCosimulation:
...
         monitor_infractions: bool = False,
-        render_birdview: bool = False,
+        get_birdview: bool = False,
+        random_seed: Optional[int] = None,
         # sufficient arguments to initialize must also be included
         **kwargs,
     ):
         self._location = location
-        response = initialize(location=location, **kwargs)
+        self.rng = random.Random(random_seed)
```
adscib, line 44: If the seed is not given, I would prefer not to use a generator at all in the client, lest the customer gets the impression that they're somehow supposed to provide seed to the API calls.

examples/npc_only.py
```diff
@@ -26,18 +26,18 @@
...
     simulation.step(current_ego_agent_states=[])
-    collision, offroad, wrong_way = simulation.infractions
+    collision, offroad, wrong_way = simulation.infraction_rates
     pbar.set_description(
-        f"Collision rate: {100*np.array(collision).mean():.2f}% | "
-        + f"Off-road rate: {100*np.array(offroad).mean():.2f}% | "
-        + f"Wrong-way rate: {100*np.array(wrong_way).mean():.2f}%"
+        f"Collision rate: {100*collision:.2f}% | "
+        + f"Off-road rate: {100*offroad:.2f}% | "
+        + f"Wrong-way rate: {100*wrong_way:.2f}%"
```
adscib, line 40: Instead of percentages, would it make more sense to print infraction rates as fractions? Something like `2/11`. It's easier to read and reconcile with the image than the percentages.

adscib (changes requested):

invertedai/cosimulation.py
```diff
@@ -98,17 +113,30 @@ def ego_states(self):
...
     @property
-    def infractions(self) -> List[InfractionIndicators]:
+    def infractions(self) -> Optional[List[InfractionIndicators]]:
         """
         If `monitor_infractions` was set in the constructor,
         lists infractions currently committed by each agent, including ego agents.
         """
         return self._infractions
 
+    def get_infraction(self) -> Optional[Tuple[float, float, float]]:
```
adscib, line 123: Can we just get rid of this function? I don't want to put anything here that's only used for our examples and we don't expect the customer to want to use. If the goal is to avoid duplication, we can have a shared module in the examples folder.

examples/npc_only.py
```diff
@@ -22,22 +22,24 @@
...
     response.birdview_image.decode_and_save(file_path)
+
 simulation = iai.BasicCosimulation(
     location=args.location,
     agent_count=10,
     monitor_infractions=True,
-    render_birdview=True,
     ego_agent_mask=[False] * 10,
+    get_birdview=True,
+    random_seed=1,
```
adscib, line 32: I would avoid specifying a random seed in the example, so that if the customer runs it multiple times they get the same result. Anyone wondering whether they can provide a seed can easily check the docs.

Ruishenl (changes requested):
What is `onp_only_notebook` for ?? I removed it on develop branch yesterday. And also `npc_only_colab.ipynb` is broken now i think, i cannot open it in this branch

adscib:
Should I review this again or is some stuff still known to be broken?

Ruishenl:
I still cannot open `npc_only_colab.ipynb`, develop works for me though

AlirezaMorsali:
> I still cannot open `npc_only_colab.ipynb`, develop works for me though

I merged develop, and didn't test the notebooks, thought it would be fine.
Let me check.

AlirezaMorsali:
All the example codes work with [this version of the server](https://github.com/inverted-ai/invertedai-drive-server/pull/42)

Ruishenl (approved):

adscib (approved):

PR #64, opened 2022-10-24
Example tweaks

adscib:
I've done some work on the CARLA example to clean it up and explain what's going on, but more is needed. It has a bunch of configuration options whose behavior is difficult to understand, so those should either be explained or removed. Generally, it's totally OK to remove some options from the code if they're not interesting from the demo perspective. We should endeavor to make this example simple and informative, as it will be the main starting point for people to start doing their own integrations.


adscib:
Since this has stalled for now, can we merge the existing changes? I think it's an improvement over what's currently there so there's no reason to hold off.

docs/source/examples/index.md
```diff
@@ -1,18 +1,12 @@
...
 
-{ref}`CARLA Example`
-
----
-```{toctree}
-:maxdepth: 2
-
-carlasim
+```{eval-rst}
+`Examples <https://github.com/inverted-ai/invertedai/tree/master/examples>`_
```
adscib, line 4: Can we make this a redirect instead?

examples/carla/carla_simulator.py
```diff
@@ -21,32 +24,38 @@
...
+    fps: int = 10  #: 10 is the only value currently compatible with Inverted AI API
+    traffic_count: int = 20  #: total number of vehicles to place in simulation
+    episode_length: int = 20  #: in seconds
+    follow_ego: bool = False  #: whether the spectator camera follows the ego vehicle
+    slack: int = 3  #: in meters, how far outside supported area to track NPCs
+    seed: float = time.time()  #: random number generator seed to control stochastic behavior
+    flag_npcs: bool = True  #: whether to display a blue dot on Inverted AI NPCs
+    flag_ego: bool = True  #: whether to display a red dot on the ego vehicle
+    ego_autopilot: bool = True  #: wheter to use traffic manager to control ego vehicle
+    npcs_autopilot: bool = False  #: ???
```
adscib, line 46: I couldn't figure out what happens if I set it to `True`. Can you either explain that here or remove this option?

examples/carla/carla_simulator.py
```diff
@@ -504,11 +581,15 @@ def _to_transform(poses: list) -> Tuple[list, list]:
...
         entrance = []
         for sp in spawn_points:
             distance = math.sqrt(
-                ((sp.location.x - self.cfg.roi_center.x) ** 2)
-                + ((sp.location.y - self.cfg.roi_center.y) ** 2)
+                ((sp.location.x - self.roi_center.x) ** 2)
+                + ((sp.location.y - self.roi_center.y) ** 2)
             )
             if (
                 self.proximity_threshold - self.cfg.slack
```
adscib, line 595: Can we handle the exits and entrances using `is_in_supported_area`? It would be very illustrative for users. We can adjust those boundaries server side if the ones defined there aren't good.

adscib (commented):
I'm done with this PR. I left a few comments about things that should be addressed but I wasn't able to.

examples/carla/carla_simulator.py
AlirezaMorsali: We can but currently, the supported_area on the server side is quite bigger than what we have on the client side, and since the z coordinate of the agents is not changed the cars may fly or go into the ground!

examples/carla/carla_simulator.py
AlirezaMorsali: If it's True, NPCs can be controlled by CARL. I had it initially to test the environment but kept it in case we need it in the future. I can either remove it or add a comment for it

adscib: probably best to remove it then

examples/carla/carla_simulator.py
adscib: I opened an issue to address this later.

AlirezaMorsali (approved):

PR #71, opened 2022-10-31
Keyboard control

Ruishenl:


AlirezaMorsali (approved):

adscib (commented):
I'm trying to run this, but downloading additional maps is expected to take multiple hours. If that's generally how long it takes, I don't think we should be including them by default.

PR #77, opened 2022-11-04
add validation

AlirezaMorsali:
I have started the code for more validation on the client side.
Wanted to check if it's in the right direction before spending more time on it.


Ruishenl:
I could be wrong but to me decorators are useful when one can be shared in multiple places or make the code structure less messy. If each decorator is only working for a specific function then why not just use a function instead

AlirezaMorsali:
> I could be wrong but to me decorators are useful when one can be shared in multiple places or make the code structure less messy. If each decorator is only working for a specific function then why not just use a function instead

I know it's syntax abuse and I had it as a function first. I thought it may be cleaner as it adds one line to the main function. I can rewrite it as a function.

invertedai/api/validations.py
```diff
@@ -0,0 +1,97 @@
...
+            assert isinstance(location, str)
+            assert all(map(lambda s: isinstance(s, AgentState), agent_states))
+            assert all(map(lambda s: isinstance(s, AgentAttributes), agent_attributes))
+            assert all(map(lambda s: isinstance(s, RecurrentState), recurrent_states))
+            assert isinstance(traffic_lights_states, (list, type(None)))
+            assert isinstance(get_birdview, bool)
+            assert isinstance(get_infractions, bool)
+            assert isinstance(random_seed, (int, type(None)))
+        except AssertionError:
+            raise InvalidInputType("iai.drive: invalid input type.")
```
adscib, line 33: Those errors should be as specific as possible, allowing the user to easily debug things.

invertedai/api/validations.py
```diff
@@ -0,0 +1,97 @@
...
+    RecurrentState,
+    AgentState,
+    AgentAttributes,
+    TrafficLightId,
+    TrafficLightState,
+)
+
+
+def drive_validation(drive):
+    def wrapped_drive(
```
adscib, line 13: This is unnecessarily complicated. There is no reuse, so you can just insert the validation code block into the relevant function (drive/initialize/etc.).

invertedai/api/validations.py
```diff
@@ -0,0 +1,97 @@
...
+    ):
+        try:
+            assert isinstance(location, str)
+            assert all(map(lambda s: isinstance(s, AgentState), agent_states))
+            assert all(map(lambda s: isinstance(s, AgentAttributes), agent_attributes))
+            assert all(map(lambda s: isinstance(s, RecurrentState), recurrent_states))
+            assert isinstance(traffic_lights_states, (list, type(None)))
+            assert isinstance(get_birdview, bool)
+            assert isinstance(get_infractions, bool)
+            assert isinstance(random_seed, (int, type(None)))
```
adscib, line 31: It's generally desirable to allow for duck typing in Python, accepting objects that have different types but the same interface. As a rule of thumb, if the rest of the function was going to execute correctly with some input and return a correct result, this input should not be rejected.

adscib (changes requested):
Good idea to ask for feedback early. Generally, what we're after here is to catch inputs that would produce some error and provide a helpful message instead, without calling the API.

Ruishenl:
what's the status of this pr ? is it ready to be merged yet ?

AlirezaMorsali:
> what's the status of this pr ? is it ready to be merged yet ?

Haven't gotten any chance to work on it, as there were more urgent issues. In particular, the traffic-light PR should be ready, just waiting for reviews.

invertedai/common.py
```diff
@@ -1,29 +1,39 @@
...
     It should not be modified, but rather passed along as received.
     """
 
-    packed: List[float] = field(default_factory=lambda: [0.0] * 132)
+    packed: List[float] = [0.0] * RECURRENT_SIZE
     #: Internal representation of the recurrent state.
 
+    @root_validator
+    @classmethod
+    def check_passwords_match(cls, values):
```
Ruishenl, line 25: This function name needs to be changed

Ruishenl (changes requested):

Ruishenl (approved):

PR #126, opened 2023-06-22
Demo updates

KieranRatcliffeInvertedAI:
Made some fixes so that the current demo's visualizations are fixed as well as adding some other code cleanup.


adscib (changes requested):
Here are some initial points to address. I'll take another look once those are taken care of.

Displaying images after each call to drive doubles the runtime of that cell on my machine, from 30 to 60 seconds, and I don't think it's worth it.

In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.

Response time notebook installs invertedai, which only makes sense in Colab, but the notebook name doesn't say that it is to be run in Colab.

The NPC only notebooks have different content in the colab vs. notebook version, which is rather confusing. Do we want to keep both those versions or only one? If the latter, at least make sure to separate the cells that install packages from those that import them, so that we don't have to run the former when using the notebook locally.

adscib:
I also adjusted the installation instructions to separate carla from other packages, since it's not needed for the notebook demos but it doesn't work with new Python versions.

KieranRatcliffeInvertedAI:
> Here are some initial points to address. I'll take another look once those are taken care of.
> 
> Displaying images after each call to drive doubles the runtime of that cell on my machine, from 30 to 60 seconds, and I don't think it's worth it.
> 
> In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.
> 
> Response time notebook installs invertedai, which only makes sense in Colab, but the notebook name doesn't say that it is to be run in Colab.
> 
> The NPC only notebooks have different content in the colab vs. notebook version, which is rather confusing. Do we want to keep both those versions or only one? If the latter, at least make sure to separate the cells that install packages from those that import them, so that we don't have to run the former when using the notebook locally.

I think we should only keep one version to keep things simple, I'm not sure why we would have separate demos for Colab and Notebook given any differences between them are not relevant for our demonstration purposes.

KieranRatcliffeInvertedAI:
> Here are some initial points to address. I'll take another look once those are taken care of.
> 
> Displaying images after each call to drive doubles the runtime of that cell on my machine, from 30 to 60 seconds, and I don't think it's worth it.
> 
> In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.
> 
> Response time notebook installs invertedai, which only makes sense in Colab, but the notebook name doesn't say that it is to be run in Colab.
> 
> The NPC only notebooks have different content in the colab vs. notebook version, which is rather confusing. Do we want to keep both those versions or only one? If the latter, at least make sure to separate the cells that install packages from those that import them, so that we don't have to run the former when using the notebook locally.

For point number 4, it seems like the only difference is that we use a Jupyter specific visualization tool in the IAI suite called Jupyter_Render. Given ScenePlotter is agnostic of platform, I don't think we should be devoting extra demos to second version of essentially the same demo (since the differences between Colab and Jupyter are negligible in this case) for what appears to be an out of date tool that also requires get_birdviews which is, as we know, a much slower visualizer.

KieranRatcliffeInvertedAI:
> Here are some initial points to address. I'll take another look once those are taken care of.
> 
> Displaying images after each call to drive doubles the runtime of that cell on my machine, from 30 to 60 seconds, and I don't think it's worth it.
> 
> In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.
> 
> Response time notebook installs invertedai, which only makes sense in Colab, but the notebook name doesn't say that it is to be run in Colab.
> 
> The NPC only notebooks have different content in the colab vs. notebook version, which is rather confusing. Do we want to keep both those versions or only one? If the latter, at least make sure to separate the cells that install packages from those that import them, so that we don't have to run the former when using the notebook locally.

I have implemented the changes for your review.

adscib:
It looks like we have multiple notebooks doing the same thing. I guess it can wait for another PR, but we should select what we want to keep. It will also be less effort for you to fix those notebooks if there are fewer of them.

> In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.

Controlled intersection notebook, and at a glance the others as well, is still not producing gifs correctly. I get one scene before running a notebook, another from INITIALIZE, and yet a different one after calling DRIVE. At a glance it looks like you're not using scene plotter to save the gif correctly.

KieranRatcliffeInvertedAI:
> It looks like we have multiple notebooks doing the same thing. I guess it can wait for another PR, but we should select what we want to keep. It will also be less effort for you to fix those notebooks if there are fewer of them.
> 
> > In the controlled intersection notebook, the gif initially in the notebook does not correspond to what you get after running the notebook. This is possibly because you don't set the random seed for DRIVE, but maybe also the initial gifs are stale.
> 
> Controlled intersection notebook, and at a glance the others as well, is still not producing gifs correctly. I get one scene before running a notebook, another from INITIALIZE, and yet a different one after calling DRIVE. At a glance it looks like you're not using scene plotter to save the gif correctly.

This is true, we probably only need to keep the master demo (IAI_demo.py) and the cosimulation demos. Maybe keep the response time demo too since it is fairly small. We may be able to keep the region_drive.py demo but this PR is only looking at the notebook demos for now.

It looks like there's been some issues with the notebooks saving after running a fresh run to get all the static output it should be fixed now.

adscib (approved):
I checked the IAI_demo notebook and it seems to be working correctly now. I'll leave it to you to decide which notebooks to remove, but let's get this thing merged already.

KieranRatcliffeInvertedAI:
Okay I have removed the region_drive.py demo since it is a duplicate in the carla examples folder and removed the Controlled_intersection.ipynb. Everything else is either unique or specifically referenced somewhere on the website.

KieranRatcliffeInvertedAI:
I deleted some other unnecessary demos and updated the python demos to use ScenePlotter. We will need to update the website to reference some changed names of files.

PR #128, opened 2023-06-24
Add blame api

zhangk1551:
Python blame api

View the notebook for detailed usage:
https://github.com/inverted-ai/invertedai/blob/5bd962be12ab75de6dd4ba0b14a2740252d027cd/examples/blame_example.ipynb

Related server api: https://github.com/inverted-ai/driving-models/pull/803


invertedai/api/mock.py
```diff
@@ -28,6 +28,18 @@ def get_mock_recurrent_state() -> RecurrentState:
     return RecurrentState()  # TODO
 
 
+def get_mock_blamed_result() -> Tuple[int, ...]:
+    return (0, 1)
```
Ruishenl, line 32: nitpick redundant brackets

Ruishenl (changes requested):
Looks fine overall, i will review again once the other updates are pushed

zhangk1551:
birdview:
part-iai-drive-side-road-green
Client API result:
image

invertedai/api/blame.py
```diff
@@ -0,0 +1,161 @@
...
+    traffic_light_state_history: Optional[List[TrafficLightStatesDict]] = None,
+    get_reasons: bool = False,
+    get_confidence_score: bool = False,
+    get_birdviews: bool = False,
+    detect_collisions: bool = False
+) -> BlameResponse:
+    """
+    Parameters
+    ----------
+
```
Ruishenl, line 51: We will need to finish filling the doc string here just like drive.py, the format we use is numpy i believe

Ruishenl (changes requested):

invertedai/api/blame.py
zhangk1551: I just copy what we have in the server side ([external.py](https://github.com/inverted-ai/driving-models/pull/803/files/ccd7a68d60e4ce1dbeb3dbf68edb8735e05b6a8c#diff-7d808c6ee389bbf93436dd303a4c8fde29cbe5992eeb27dd1d3643f757aca9daR119)) to here. Once the description there is updated, the doc here need to be updated accordingly.

invertedai/api/blame.py
```diff
@@ -48,6 +48,41 @@ def blame(
     """
     Parameters
     ----------
+    location:
```
Ruishenl, line 51: the variable uses `get_birdviews` but in the docstring is `get_birdview`; and `detect_collisions`  is missing in the docstring

zhangk1551: Because `detect_collisions` is only for internal usage, I skipped it here, so it won't occur in the public doc.

invertedai/api/blame.py
Ruishenl, line 1: we don't want to include detect_collisions in the doc?

zhangk1551: Yes. The external users are not supposed to use this flag.

zhangk1551: It's used in the `examples`. I assumed that there is a stage called "internal testing" according to @adscib , this is when this flag can be used.

Ruishenl: I see but they can still view the source code .   I think this is fine for now if we are okay  if they find out and use this option

zhangk1551: I guess if it's not in the doc, that means we don't guarantee to support `detect_collisions` formally. 

I think maybe that's fine for now, because we haven't publicly announced the `blame` feature. cc @adscib

Ruishenl (approved):

PR #136, opened 2023-08-15
Add blame cpp api

zhangk1551:
The cpp documentation (i.e. comments in the code) is not ready yet.

One tricky thing is that `json` only allow `string` as key, and in the Python/C++ api, we currently regard the key of the `reasons` as `int`, which is different from the `string` in the REST api.

image

### DIFF
--- a/invertedai_cpp/invertedai/api.h
+++ b/invertedai_cpp/invertedai/api.h
@@ -23,6 +23,7 @@ namespace invertedai {
  * @return the location_info response receive from the API
  * @see invertedai::initialize
  * @see invertedai::drive
+ * @see invertedai::blame
  */
 LocationInfoResponse location_info(LocationInfoRequest &location_info_request,
                                    Session *session);
@@ -42,6 +43,7 @@ LocationInfoResponse location_info(LocationInfoRequest &location_info_request,
  * @return the initialize response receive from the API
  * @see invertedai::location_info
  * @see invertedai::drive
+ * @see invertedai::blame
  */
 InitializeResponse initialize(InitializeRequest &initialize_request,
                               Session *session);
@@ -50,24 +52,26 @@ InitializeResponse initialize(InitializeRequest &initialize_request,
  * Wrap the REST API "drive".
  * Drive the agents based on given situations.
  *
- * @param drive_request the initialize request to send to the API
+ * @param drive_request the drive request to send to the API
  * @param session the shared session connected with the host
- * @return the initialize response receive from the API
+ * @return the drive response receive from the API
  * @see invertedai::location_info
  * @see invertedai::initialize
+ * @see invertedai::blame
 */
 DriveResponse drive(DriveRequest &drive_request, Session *session);
 
 
 /**
- * Wrap the REST API "drive".
- * Drive the agents based on given situations.
+ * Wrap the REST API "blame".
+ * Blame the agents at fault in the collision.
  *
- * @param drive_request the initialize request to send to the API
+ * @param blame_request the blame request to send to the API
  * @param session the shared session connected with the host
- * @return the initialize response receive from the API
+ * @return the blame response receive from the API
  * @see invertedai::location_info
  * @see invertedai::initialize
+ * @see invertedai::drive
 */
 BlameResponse blame(BlameRequest &blame_request, Session *session);
 
--- a/invertedai_cpp/invertedai/api.h
+++ b/invertedai_cpp/invertedai/api.h
@@ -2,6 +2,8 @@
  * Interface wrappers for the REST API.
  */
 
+#include "blame_request.h"
+#include "blame_response.h"
 #include "drive_request.h"
 #include "drive_response.h"
 #include "initialize_request.h"
@@ -55,5 +57,18 @@ InitializeResponse initialize(InitializeRequest &initialize_request,
  * @see invertedai::initialize
 */
 DriveResponse drive(DriveRequest &drive_request, Session *session);
+
+
+/**
+ * Wrap the REST API "drive".
+ * Drive the agents based on given situations.
+ *
+ * @param drive_request the initialize request to send to the API
+ * @param session the shared session connected with the host
+ * @return the initialize response receive from the API
+ * @see invertedai::location_info
+ * @see invertedai::initialize
+*/
+BlameResponse blame(BlameRequest &blame_request, Session *session);
 
 } // namespace invertedai
--- a/invertedai_cpp/invertedai/blame_request.cc
+++ b/invertedai_cpp/invertedai/blame_request.cc
@@ -28,15 +28,21 @@ BlameRequest::BlameRequest(const std::string &body_str) {
     AgentAttributes agent_attribute = {element[0], element[1], element[2]};
     this->agent_attributes_.push_back(agent_attribute);
   }
-  this->traffic_light_state_history_.clear();
-  for (const auto &elements : this->body_json_["traffic_light_state_history"]) {
-    std::vector<TrafficLightState> traffic_light_states;
-    traffic_light_states.clear();
-    for (const auto &element : elements) {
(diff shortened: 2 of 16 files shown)


invertedai_cpp/invertedai/blame_request.cc
```diff
@@ -0,0 +1,169 @@
...
+std::string BlameRequest::body_str() {
+  this->refresh_body_json_();
+  return this->body_json_.dump();
+}
+
+std::string BlameRequest::location() const { return this->location_; }
+
+std::pair<int, int> BlameRequest::colliding_agents() const {
+  return this->colliding_agents_;
+}
```
KieranRatcliffeInvertedAI, line 110: Would it possibly be wise to make this a "std::vector" for future proofing reasons in case we want to be more flexible with the number of agents involved in a collision or do you think this is unnecessary?

zhangk1551: I want to keep consistent with the Python API. We can change it as needed in the future.
image

invertedai_cpp/invertedai/blame_request.cc
```diff
@@ -0,0 +1,169 @@
...
+      std::vector<TrafficLightState> traffic_light_states;
+      traffic_light_states.clear();
+      for (const auto &element : elements) {
+        TrafficLightState traffic_light_state = {element[0], element[1]};
+        traffic_light_states.push_back(traffic_light_state);
+      }
+      traffic_light_state_history.push_back(traffic_light_states);
+    }
+    this->traffic_light_state_history_ = traffic_light_state_history;
+  }
```
KieranRatcliffeInvertedAI, line 46: I know we use the "auto" keyword in the other request.cc files and this isn't exclusive to this PR in particular but is there a reason we don't use explicit type definitions instead? Is it not guaranteed, for example, that **this->body_json_["agent_state_history"]** returns **std::vector<AgentState>**?

zhangk1551:
Not sure why I can't see the above comment in the github UI. Anyway, let me
reply through the email.

The body_json_ is from the library "nlohmann/json" and the types of the
class members, like "AgentState" are self-defined, which can't be converted
automatically by C++.

KieranRatcliffeInvertedAI:
Thanks for the info, I deleted the comment since I came to the same
conclusion. I should be done combing through the PR soon.

Cheers,


Kieran Ratcliffe
Customer Experience Engineer - Artificial Intelligence
M.Sc. Mechanical Engineering - McGill University
B.A.Sc. Systems Design Engineering - University of Waterloo

invertedai_cpp/invertedai/blame_response.cc
```diff
@@ -0,0 +1,105 @@
...
+}
+
+std::string BlameResponse::body_str() {
+  this->refresh_body_json_();
+  return this->body_json_.dump();
+}
+
+std::vector<int> BlameResponse::agents_at_fault() const {
+  return this->agents_at_fault_;
+}
```
KieranRatcliffeInvertedAI, line 70: Just to confirm, agents_at_fault is NOT implemented as the type std::optional<...> because there is no flag option to turn off this value like reasons, confidence_score, and birdviews, correct? If that is the reason I'm wondering why we don't implement std::optional<...> for similar parameters in our other C++ response endpoints. Perhaps this is a matter for another PR if std::optional<...> is better.

KieranRatcliffeInvertedAI:
I've run the example which appears to run well and the rest of the C++ SDK looks good to me.

KieranRatcliffeInvertedAI (approved):

invertedai_cpp/invertedai/api.h
```diff
@@ -56,4 +58,17 @@ InitializeResponse initialize(InitializeRequest &initialize_request,
...
+ * Drive the agents based on given situations.
+ *
+ * @param drive_request the initialize request to send to the API
+ * @param session the shared session connected with the host
+ * @return the initialize response receive from the API
+ * @see invertedai::location_info
+ * @see invertedai::initialize
+*/
+BlameResponse blame(BlameRequest &blame_request, Session *session);
+
```
KieranRatcliffeInvertedAI, line 73: Update the documentation here to reference blame.

invertedai_cpp/invertedai/blame_response.h
```diff
@@ -0,0 +1,77 @@
...
+  BlameResponse(const std::string &body_str);
+  /**
+   * Serialize all the fields into a string.
+   */
+  std::string body_str();
+
+  // getters
+  /**
+   * Get a tuple containing all agents predicted to be at fault. If empty, BLAME
+   * has predicated no agents are at fault.
```
KieranRatcliffeInvertedAI, line 34: Update this documentation to lose references to python variable types (tuple etc.)

invertedai_cpp/invertedai/blame_response.h
```diff
@@ -0,0 +1,77 @@
...
+  std::vector<int> agents_at_fault() const;
+  /**
+   * Get float value between [0,1] indicating BLAME’s confidence in the response
+   * where 0.0 represents the minimum confidence and 1.0 represents maximum.
+   */
+  std::optional<float> confidence_score() const;
+  /**
+   * Get a dictionary with agent IDs as keys corresponding to “agents_at_fault”
+   * paired with a list of reasons why the keyed agent is at fault (e.g.
+   * traffic_light_violation).
```
KieranRatcliffeInvertedAI, line 45: Similar here, remove reference to python types.

invertedai_cpp/invertedai/blame_response.h
```diff
@@ -0,0 +1,77 @@
...
+   */
+  std::optional<std::map<int, std::vector<std::string>>> reasons() const;
+  /**
+   * Get the images visualizing the collision case.
+   */
+  std::optional<std::vector<std::vector<unsigned char>>> birdviews() const;
+
+  // setters
+  /**
+   * Set a tuple containing all agents at fault.
```
KieranRatcliffeInvertedAI, line 55: Change reference to python tuple data type.

invertedai_cpp/invertedai/blame_response.h
```diff
@@ -0,0 +1,77 @@
...
+   */
+  void set_agents_at_fault(const std::vector<int> &agents_at_fault);
+  /**
+   * Set float value between [0,1] indicating BLAME’s confidence in the response
+   */
+  void set_confidence_score(std::optional<float> confidence_score);
+  /**
+   * Set a dictionary with agent IDs as keys corresponding to “agents_at_fault”
+   * paired with a list of reasons why the keyed agent is at fault (e.g.
+   * traffic_light_violation).
```
KieranRatcliffeInvertedAI, line 65: Update reference to python dictionary data type.

KieranRatcliffeInvertedAI (changes requested):

invertedai_cpp/invertedai/blame_response.cc
zhangk1551: Yeah, currently the `optional` feature in the C++ responses is not strictly taken care of. I tried the REST api and found that even though when the `get_***` flags are turned off, the REST responses will contain default values for the corresponding fields instead of returning null, and thus won't cause issues to the C++ response parsing. I agree that we can wrap the fields more clearly. I will create another PR to address it.

KieranRatcliffeInvertedAI (approved):

KieranRatcliffeInvertedAI:
The comment updates look good to me.

KieranRatcliffeInvertedAI:
You may merge this branch @zhangk1551.

zhangk1551:
@Ruishenl This PR is merged into `master`. Not sure whether I need to re-merge it to `develop`. (The cpp api is standalone from the Python api & demos, and the PR didn't change the existing Python/C++ api except some doc typos.)

PR #143, opened 2023-11-22
Fixed a link in README pointing to correct Colab notebook

KieranRatcliffeInvertedAI:


adscib (approved):
The fix makes sense, but I'm a little confused about what's going on with branches here. So develop is not protected but documentation is?

PR #144, opened 2023-11-23
Add agent type

rf-ivtdai:
client-side update to accommodate adding agent_type to agent attributes


invertedai/common.py
```diff
@@ -141,22 +141,35 @@ class AgentAttributes(BaseModel):
...
     """
 
-    length: float  #: Longitudinal extent of the agent, in meters.
-    width: float  #: Lateral extent of the agent, in meters.
+    length: Optional[float] = None  #: Longitudinal extent of the agent, in meters.
+    width: Optional[float] = None  #: Lateral extent of the agent, in meters.
     #: Distance from the agent's center to its rear axis in meters. Determines motion constraints.
-    rear_axis_offset: float
+    rear_axis_offset: Optional[float] = None
+    agent_type: Optional[str] = 'car'  #: Type of agent, default to car, can also be pedestrian
```
Ruishenl, line 148: Can you make the type here the Enum instead of str ?

adscib: I'm actually not sure that an enum is preferable. It would force people to update the package whenever we introduce additional types.

invertedai/common.py
```diff
@@ -49,7 +49,7 @@ def fromlist(cls, l):
         return cls(x=x, y=y)
 
     def __sub__(self, other):
-        return math.sqrt((abs(self.x - other.x)**2) + (abs(self.y - other.y)**2))
+        return math.sqrt((abs(self.x - other.x) ** 2) + (abs(self.y - other.y) ** 2))
```
adscib, line 52: `abs` is redundant when taking a square.

invertedai/common.py
```diff
@@ -141,22 +141,35 @@ class AgentAttributes(BaseModel):
...
 
     @classmethod
     def fromlist(cls, l):
-        length, width, rear_axis_offset = l
-        return cls(length=length, width=width, rear_axis_offset=rear_axis_offset)
+        if len(l) == 4:
+            length, width, rear_axis_offset, agent_type = l
+            return cls(length=length, width=width, rear_axis_offset=rear_axis_offset, agent_type=agent_type)
+        else:
+            agent_type = l[0]
```
adscib, line 156: If you're asserting that `l` only has one element, `agent_type, = l` would be preferable. If it can have more, I'm a little worried about how you know that the agent type comes first.

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+        return all([type(field) is float and field >= 0 for field in agent.tolist()[:3]])
+    elif agent.agent_type == AgentType.pedestrian.value:
+        return all([type(field) is float and field >= 0 for field in agent.tolist()[:2]])
+    else:
+        return False
+
+
+def validate_attributes_without_history(agent_attributes: List[AgentAttributes]):
+    if agent_attributes is not None:
+        if not all([len(agent.tolist()) == 1 and type(agent.agent_type) is str for agent in agent_attributes]):
```
adscib, line 17: Is it actually illegal to use `agent_type=None` here?

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+            raise InvalidInputType(
+                "agent_attributes: cannot specify this combination of attributes without full state history.")
+
+
+def validate_attributes_with_history(states_history: List[List[AgentState]], agent_attributes: List[AgentAttributes],
+                                     agent_count: int):
+    if agent_attributes is None:
```
adscib, line 24: This is not allowed under the type signature.

invertedai/common.py
Ruishenl: If we are using string enums it will be automatically casted as string and when we introduce new types we just append more values. Plus it is  using enums for validation already on client side

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+            "agent attributes: need to specify length, width, rear_axis_offsets given full state history.")
+    else:
+        if len(states_history[0]) != len(agent_attributes):
+            raise InvalidInput(
+                f"The number of agents in agents_attributes({len(agent_attributes)}) does not "
+                f"match the number of agents in states_history({len(states_history[0])})."
+            )
+        if len(states_history[0]) > agent_count:
+            raise InvalidInput(
+                f"num_agents_to_spawn ({agent_count}) must be larger than the number of conditional agents"
```
adscib, line 35: I guess "not smaller"?

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+                "agent_attributes: cannot specify this combination of attributes without full state history.")
+
+
+def validate_attributes_with_history(states_history: List[List[AgentState]], agent_attributes: List[AgentAttributes],
+                                     agent_count: int):
+    if agent_attributes is None:
+        raise InvalidInputType(
+            "agent attributes: need to specify length, width, rear_axis_offsets given full state history.")
+    else:
+        if len(states_history[0]) != len(agent_attributes):
```
adscib, line 28: Should we also check that all lists in states_history have the same length?

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+
+def validate_drive_flows(agent_states: List[AgentState],
+                         agent_attributes: List[AgentAttributes],
+                         recurrent_states: List[RecurrentState]):
+    if len(agent_states) != len(agent_attributes):
+        raise InvalidInput("Incompatible Number of Agents in either 'agent_states' or 'agent_attributes'.")
+    if len(agent_states) != len(recurrent_states):
+        raise InvalidInput("Incompatible Number of Agents in either 'agent_states' or 'recurrent_states'.")
+    if not all([agent.agent_type == AgentType.car.value for agent in agent_attributes]):
+        raise InvalidInput("unsupported agent type in agents_attributes.")
```
adscib, line 62: Here and throughout, the more information you can include in the error message, the better. For example, you can specify the lengths of both lists, and the first unsupported agent type found.

adscib (changes requested):
Generally looks good. I left some comments to make sure that the validators match the intended semantics.

We should also update documentation throughout. At the very least, update the user guide to indicate that we support pedestrians, and the main docstring for DRIVE to warn that it doesn't currently support pedestrians, but will in the future.

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+                f"match the number of agents in states_history({len(states_history[0])})."
+            )
+        if len(states_history[0]) > agent_count:
+            raise InvalidInput(
+                f"num_agents_to_spawn ({agent_count}) must be larger than the number of conditional agents"
+                f" passed ({len(states_history[0])})."
+            )
+        if not all([all_float_fields_valid(agent) for agent in agent_attributes]):
+            raise InvalidInput(
+                "agent attributes: need to specify length, width, rear_axis_offsets in valid range,"
```
Ruishenl, line 40: We should mention pedestrian doesn't have lr in the error messages here

invertedai/api/validations.py
```diff
@@ -0,0 +1,62 @@
...
+from invertedai.common import AgentAttributes, AgentState, RecurrentState, AgentType
+
+
+def all_float_fields_valid(agent: AgentAttributes):
+    if agent.agent_type == AgentType.car.value:
+        return all([type(field) is float and field >= 0 for field in agent.tolist()[:3]])
+    elif agent.agent_type == AgentType.pedestrian.value:
+        return all([type(field) is float and field >= 0 for field in agent.tolist()[:2]])
+    else:
+        return False
```
Ruishenl, line 12: Maybe we should raise a type not supported exception here ?

invertedai/common.py
Ruishenl, line 156: It's probably safer to do agent_type = l[-1]

rf-ivtdai:
I have addressed the comments and documented the new features in API and class docstrings. I am a bit unsure as to whether we want to keep the client-side request validations for initialize and drive. After testing, the current checks are all handled gracefully by the server. Replicating the same logic here is redundant and error-prone.

invertedai/common.py
rf-ivtdai: keeping string as per our conversation.

invertedai/common.py
```diff
@@ -141,22 +141,34 @@ class AgentAttributes(BaseModel):
...
     """
 
-    length: float  #: Longitudinal extent of the agent, in meters.
-    width: float  #: Lateral extent of the agent, in meters.
+    length: Optional[float] = None  #: Longitudinal extent of the agent, in meters.
+    width: Optional[float] = None  #: Lateral extent of the agent, in meters.
     #: Distance from the agent's center to its rear axis in meters. Determines motion constraints.
-    rear_axis_offset: float
+    rear_axis_offset: Optional[float] = None
+    agent_type: Optional[str] = 'car'  #: Type of agent, default to car, can also be pedestrian
```
adscib, line 148:
```suggestion
    agent_type: Optional[str] = 'car'  #: Valid types are those in `AgentType`, but we use `str` here for extensibility.
```
I actually liked having an `AgentType` enum, even though we use `str` here. We need to have an authoritative list of valid agent types somewhere, and it only makes sense that we would expose it programmatically. Let's just be clear that this enum is extensible and we expect it to grow in future releases.

adscib (commented):
I updated the docstring, but please make sure it is accurate.

rf-ivtdai:
To better measure alignments between client and server, i'll add regression tests that mirrors the server integration tests for initialize and drive

KieranRatcliffeInvertedAI:
Changes to the C++ SDK are complete, @adscib is there anything else that needs to be changed before we can merge?

rf-ivtdai:
Regression tests added:
initialize: 10 positive flow - expect 200
                  5 negative flow - expect 400
drive: 4 positive flow - expect 200
drive: 1 negative flow - expect 400
These are all relevant to add_agent_types feature, all passing against local server

KieranRatcliffeInvertedAI:
@adscib I know @rf-ivtdai has been talking to me about adding regression tests which he seems to have added some on the python SDK. Should we add some to C++ SDK or should that be a different PR?

adscib:
It would be nice to add some tests, but I don't want to delay this release, so a separate PR please. Are we waiting for anything here, or can we release already?

rf-ivtdai:
We are ready to release!

PR #151, opened 2023-12-01
Updated INITIALIZE documentation for clarity on usage of agent_attributes

KieranRatcliffeInvertedAI:


invertedai/api/initialize.py
```diff
@@ -60,17 +60,20 @@ def initialize(
...
-    specification, but the scene will be initialized as if the traffic lights were disabled.
-    Every simulation needs to start with a call to this function in order to obtain correct recurrent states for :func:`drive`.
+    Initializes a simulation in a given location, using a combination of **user-defined** and **sampled** agents.
+    **User-defined** agents are placed in a scene first, after which a number of agents are sampled conditionally 
+    inferred from the `agent_count` argument.
+    If **user-defined** agents are desired, `states_history` must contain a list of `AgentState's` of all **user-defined** 
+    agents per historical time step.
+    Any **user-defined** agent must have a corresponding fully specified static `AgentAttribute` in `agent_attributes`. 
+    Any **sampled** agents not specified in `agent_attributes` will be generated with default static attribute values however
+     **sampled** agents may be defined by specifying all static attributes or by specifying `agent_type` only.
```
rf-ivtdai, line 70: Sampled agent cannot be defined by specifying all static attributes. They can only be defined with agent_type only, or as an empty list (empty attributes).

KieranRatcliffeInvertedAI: Ah I see my mistake, I suppose the option in the common.py code for all attributes is for user-defined agents only?

KieranRatcliffeInvertedAI: If that is the case, will the back end throw a descriptive error for this specific error case?

rf-ivtdai: You will get 400 Bad request -  "agent attributes: need to specify all static attributes for agents with state history, can specify at most agent type for agents without state history."

PR #153, opened 2023-12-01
Removed an incorrect method for initializing sampled agents in the do…

KieranRatcliffeInvertedAI:
…cumentation.

@rf-ivtdai does this look good and accurate?


rf-ivtdai:
Looks good!

PR #154, opened 2023-12-05
Removed conditional_agents and agent_count parameters from initialize…

KieranRatcliffeInvertedAI:
…_request

### DIFF
--- a/invertedai_cpp/invertedai/initialize_request.cc
+++ b/invertedai_cpp/invertedai/initialize_request.cc
@@ -5,33 +5,7 @@ using json = nlohmann::json;
 namespace invertedai {
 InitializeRequest::InitializeRequest(const std::string &body_str) {
   this->body_json_ = json::parse(body_str);
-
   this->location_ = this->body_json_["location"];
-  this->num_agents_to_spawn_ = this->body_json_["num_agents_to_spawn"].is_number_integer()
-    ? this->body_json_["num_agents_to_spawn"].get<int>()
-    : 0;
-  this->conditional_agent_states_.clear();
-  for (const auto &element : this->body_json_["conditional_agent_states"]) {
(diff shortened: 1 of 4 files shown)


invertedai_cpp/invertedai/initialize_request.cc
```diff
@@ -79,8 +53,8 @@ InitializeRequest::InitializeRequest(const std::string &body_str) {
   this->get_infractions_ = this->body_json_["get_infractions"].is_boolean()
     ? this->body_json_["get_infractions"].get<bool>()
     : false;
-  this->agent_count_ = this->body_json_["agent_count"].is_number_integer()
-    ? std::optional<int>{this->body_json_["agent_count"].get<int>()}
+  this->num_agents_to_spawn_ = this->body_json_["num_agents_to_spawn"].is_number_integer()
+    ? std::optional<int>{this->body_json_["num_agents_to_spawn"].get<int>()}
```
rf-ivtdai, line 57: Thanks! As testing for this change, could you please add a dedicated test flow to initialize where num_agent_to_spawn is not set,  and both agent attributes and state history has a valid, positive size?

KieranRatcliffeInvertedAI: While I think these tests are valuable, do you think we should make that into a separate PR?

rf-ivtdai: Sure :)

KieranRatcliffeInvertedAI: Is the PR acceptable otherwise?

rf-ivtdai: As long as relevant testing will be added in the near future, yes!

PR #157, opened 2023-12-08
Update api key handling with optionally passed key type and url

Ruishenl:


rf-ivtdai (commented):
Try to refactor and avoid the continuous stream of if statements, if some of the clauses are logically connected, they should be organized so, if certain clauses overrules others, maybe that other clause shouldn't be run at all. Having this many independent if checks obfuscates the logic and makes the code harder to follow and maintain

rf-ivtdai (approved):
Keep calm and embrace changes

PR #171, opened 2024-01-16
Develop

rf-ivtdai:
Need to consolidate a few more changes before merging:

- [x] Doc string updates

- [x] Python client code update


KieranRatcliffeInvertedAI:
As discussed in person vis-a-vis traffic light documentation:
1. Add a brief note (in both python and C++ clients) indicating that LightRecurrentState pertains to a light group (e.g. an intersection) whereas TrafficLightStatesDict is per individual light.
2. Add another note (in both python and C++ clients) indicating light states passed in through TrafficLightStatesDict are prioritized over the LightRecurrentState. In the situation a user wants to manually change individual traffic lights, this will let them know LightRecurrentState does not need to be modified. My suggestion for wording would be "If manual control of individual traffic lights is desired, modify the relevant state(s) in TrafficLightStatesDict and pass in LightRecurrentState as usual."

PR #174, opened 2024-01-22
light endpoint deprecation

rf-ivtdai:
Update documentation for deprecated light endpoint, update tests and released notebooks for compatibility.
Client side: light method still available, doc string has "DEPRECATED" line, server will return 404 when called.


Ruishenl:
The code example also needs an update here https://github.com/inverted-ai/invertedai/blob/master/README.md

Ruishenl:
Also in `IAI_full_demo.ipynb` we should probably remove the "LIGHT" from  "This demo will showcase some of the capabilities available in the Inverted AI API, including INITIALIZE, DRIVE, and LIGHT."

rf-ivtdai:
> Also in `IAI_full_demo.ipynb` we should probably remove the "LIGHT" from "This demo will showcase some of the capabilities available in the Inverted AI API, including INITIALIZE, DRIVE, and LIGHT."

On this branch, the LIGHT section is already replaced with a flow that uses light recurrent states in that notebook

Ruishenl (approved):

PR #190, opened 2024-03-13
Large map utilities

KieranRatcliffeInvertedAI:
New area initialization utility that should reduce time on large maps by skipping empty space. Furthermore, added a capability to include conditional agents like regular initialize can (this last feature requires a lot of work around with how area_initialize() works for example a filter for agents at the edge of regions cannot filter out the conditional agents).

Among regular feedback, name suggestions of functions and variables would be appreciated (e.g. what should the function be called?) and should we delete the old area_initialize() function? In regards to this last question, setting the scaling_factor parameter to 0.0 will essentially give you the same output as the old area_initialize().


invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+
+    return agent_density_list
+
+def _calculate_agent_density_max_scaled(agent_density,scaling_factor,drivable_ratio,max_drivable_area_ratio):
+
+    return round(agent_density*(1-scaling_factor*(max_drivable_area_ratio-drivable_ratio)/max_drivable_area_ratio)) if drivable_ratio > 0.0 else 0
+
+def fill_initialization(
+    location, 
+    agent_density,
```
rf-ivtdai, line 632: Same as above.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+    map_center=(0, 0),
+    width=100, 
+    height=100, 
+    stride=50, 
+    initialize_fov=INITIALIZE_FOV, 
+    get_birdview=False,
+    birdview_path=None,
+    conditional_agent_attributes=[],
+    conditional_agent_states=[[]],
+    scaling_factor = 1.0,
```
rf-ivtdai, line 644: Another option is to insert  """ """ doctrings for these arguments and explain what they mean and provide a simple example.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+
+    centers = get_centers(
+        map_center, 
+        height, 
+        width, 
+        stride
+    )
+
+    agent_density_list = _get_drivable_area_ratio(centers,location,agent_density,scaling_factor)
+
```
rf-ivtdai, line 663: _get_drivable_area_ratio returns a list of "agent count by grid center", so the method should probably be renamed as such?

invertedai/utils.py
KieranRatcliffeInvertedAI: Good idea.

invertedai/utils.py
KieranRatcliffeInvertedAI: I agree agent density is not explanatory but I would like this function to be backwards compatible for the customers who already user the area_initilization function. Unless we are keeping both of these functions, in which case sure I am happy to change that parameter name to something more intuitive.

invertedai/utils.py
KieranRatcliffeInvertedAI: That makes sense.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+def fill_initialization(
+    location, 
+    agent_density, 
+    traffic_lights_states=None, 
+    random_seed=None, 
+    map_center=(0, 0),
+    width=100, 
+    height=100, 
+    stride=50, 
+    initialize_fov=INITIALIZE_FOV,
```
adscib, line 639: It's weird to me that this is configurable, given that `initialize` itself does not allow fov to be changed.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+        # Remove all conditional agents before filtering at the edges
+        non_conditional_agent_states = response.agent_states[len(states_history_specified_area):]
+        non_conditional_agent_attributes = response.agent_attributes[len(agent_attributes_specified_area):]
+        non_conditional_recurrent_states = response.recurrent_states[len(states_history_specified_area):]
+
+        # Filter out agents that are not inside the ROI to avoid collision with other agents not passed as conditional
+        # SLACK is for removing the agents that are very close to the boundary and
+        # they may collide agents not filtered as conditional
+        valid_agents = list(filter(
+            lambda x: inside_fov(center=area_center, initialize_fov=initialize_fov - SLACK, point=x[0].center),
```
adscib, line 755:
```suggestion
            lambda x: inside_fov(center=area_center, initialize_fov=min(2*stride, initialize_fov - SLACK), point=x[0].center),
```
Based on the assumption that INITIALIZE works better near the center, where it has more context (maybe @bzwartsenberg can speak to whether that's true) it would make sense to crop tighter and let each region be primarily initialized by the call with the closest center.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+                continue
+
+        for _ in range(1):
+            try:
+                # Initialize simulation with an API call
+                response = iai.initialize(
+                    location=location,
+                    states_history=states_history,
+                    agent_attributes=agent_attributes_con,
+                    agent_count=agent_density_area_center,
```
adscib, line 724: This argument is deprecated in initialize, so it would be better to pad `agent_attributes` instead. That would also avoid the confusing case where `agent_count` is smaller than the number of conditional agents passed in.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+    agent_density, 
+    traffic_lights_states=None, 
+    random_seed=None, 
+    map_center=(0, 0),
+    width=100, 
+    height=100, 
+    stride=50, 
+    initialize_fov=INITIALIZE_FOV, 
+    get_birdview=False,
+    birdview_path=None,
```
adscib, line 641: I think `get_birdview` is redundant, since the only sensible configurations are when `get_birdview == birdview_path is not None`. I would also rename `birdview_path` to `save_birdviews_to`.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+        from cv2 import (
+            cvtColor,
+            COLOR_BGR2GRAY
+        )
+
+    except ImportError as e:
+        iai.logger.error(
+            "Calculating drivable area requires cv2, which was not found."
+        )
+        raise e
```
adscib, line 589: I don't understand why this is needed at all. It seems like we're only using it to count the black pixels, which are zeros in any representation, so why do we need to import a color converter from this heavy package that's otherwise not our dependency?

adscib (changes requested):
Very nice. This will make area initialize much more useful. I left some comments related to your changes and some related to the existing code, which I think it's a good opportunity to correct. I would just remove the previous implementation in favor of this one and keep the old name. I agree with Frederic's comments regarding naming.

If we decide to break backwards compatibility here, which I think is tolerable, I would move those functions to another module. `utils` is typically designated for things that would ideally be available in other packages, not the functionality provided by this package. Since we anticipate adding more code for handling large maps, maybe we can add a module like `helpers.large_maps`?

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+        if len(agent_attributes_specified_area) < 1:
+            # Only if there are no conditional agents in the area can it be skipped.
+            if agent_density_area_center < 1:
+                #Skip if no agents are requested (e.g. in areas with no drivable surfaces)
+                continue
+            elif len(con_agent_state) >= agent_density_area_center:
+                #Skip if the calculated number of agents has been satisfied or surpassed
+                continue
+
+        for _ in range(1):
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: I'm not sure why it was included in the original area_initialize() but I don't see any reason we would run initialize() multiple times per area so it is now just a try-except.

invertedai/utils.py
KieranRatcliffeInvertedAI: It looks like the "initialize_fov" parameter isn't actually passed into the initialize() in any way, it looks like it's used to determine which existing vehicles are to be considered as conditional for the current area. So to me, this is an issue of an unclear parameter name.

invertedai/utils.py
KieranRatcliffeInvertedAI: Why would we allow the region in which the function considers conditional agents to be potentially limited by the user setting the stride parameter? For example, if the user sets the parameter to 25, initialize() will potentially produce collisions within the outer limit of the region.

KieranRatcliffeInvertedAI: Can you elaborate what you mean by "near the center", near the center of what specifically? Near the center of an arbitrary 100x100m space or near the center of an entire map?

invertedai/utils.py
KieranRatcliffeInvertedAI: Something similar is done in the Image object in common.py so I figured this was the acceptable way to do so. I can find a lighter-weight colour-to-grayscale converter or just do it manually in the code.

adscib: I see, so you can't decode the image at all without cv2. I guess we can leave this here then.

invertedai/utils.py
KieranRatcliffeInvertedAI: Added the docstring.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+
+    conditional_agent_recurrent_state_dict = {}
+    for agent in conditional_agent_states[-1]:
+        conditional_agent_recurrent_state_dict[(agent.center.x,agent.center.y)] = None
+
+    for i, area_center in enumerate(tmap(
+        Point.fromlist, 
+        centers, 
+        total=len(centers),
+        desc=f"Initializing {location.split(':')[1]}"
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+    width=100, 
+    height=100, 
+    stride=50, 
+    initialize_fov=INITIALIZE_FOV, 
+    get_birdview=False,
+    birdview_path=None,
+    conditional_agent_attributes=[],
+    conditional_agent_states=[[]],
+    scaling_factor = 1.0,
+):
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+    random_seed=None, 
+    map_center=(0, 0),
+    width=100, 
+    height=100, 
+    stride=50, 
+    initialize_fov=INITIALIZE_FOV, 
+    get_birdview=False,
+    birdview_path=None,
+    conditional_agent_attributes=[],
+    conditional_agent_states=[[]],
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Changed it to None but it is very convenient for these to be empty lists so I added an IF statement to change them back into that form but left it by default as None.

invertedai/utils.py
```diff
@@ -563,6 +575,211 @@ def inside_fov(center: Point, initialize_fov: float, point: Point) -> bool:
...
+                    get_birdview=get_birdview,
+                )
+
+                if traffic_lights_states is None:
+                    # If no traffic light states are given, take the first non-None traffic light states output as the consistent traffic light states across all areas
+                    traffic_lights_states = [response.traffic_lights_states]
+
+                break
+            except BaseException as e:
+                print(e)
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Updated this to be catch a `InvertedAIError` instead. Pydantic seems to have an issue accepting a `logger` type so I use the built-in function of the `logging` module to log exception. I made this particular exception a "warning" because it's not typically a serious thing if one region fails to sample agents.

invertedai/utils.py
KieranRatcliffeInvertedAI: Something similar is done in the Image object in common.py so I figured this was the acceptable way to do so. I can find a lighter-weight colour-to-grayscale converter or just do it manually in the code.

invertedai/utils.py
KieranRatcliffeInvertedAI: Done.

KieranRatcliffeInvertedAI: This is definitely breaking more the backwards compatibility so I think we just commit to this.

invertedai/utils.py
KieranRatcliffeInvertedAI: I added a manual grayscale conversion using Luminosity method and added a comment saying as much for future reference.

invertedai/utils.py
KieranRatcliffeInvertedAI: I think I understand what you mean. A vehicle should only be considered valid if the closest region center was the region of interest from which it was sampled, correct? I think that could represent some significant changes to how area_initialize() works beyond tuning the SLACK value (e.g. how do we find/deal with regions on the edge of drivable areas? If we prune too many agents, how will that impact remaining agents behaviour?). My opinion is that it should be a PR for another time since my best guess is that it will not impact the user interface and can be a performance improvement we roll out quietly at a later date.

rf-ivtdai (approved):

PR #197, opened 2024-03-22
Large map utilities

KieranRatcliffeInvertedAI:
Some minor updates to AreaDriver including functioning in-simulation visualization.


Ruishenl (changes requested):
It looks like the extra tab spaces you added here is not align with the core package anymore. Can you remove the extra tab space ? I think the standard is 4 spaces. it also makes things easier to review

KieranRatcliffeInvertedAI:
I've made the change back to spaces from tabs. I've also added a change so that the quadtree splitting condition is not just if the number of agents within the leaf node is above the capacity, but it also includes a buffer zone around the leaf node. This is as opposed to checking a bubble around each agent for neighbouring agents which should save execution time and allow for easier tuning of agent density.

invertedai/utils.py
```diff
@@ -445,21 +445,13 @@ def _interpret_response_line(self, result):
 
 def _get_centers(map_center, height, width, stride):
     def check_valid_center(center):
-        return (map_center[0] - width) < center[0] < (map_center[0] + width) and (
-            map_center[1] - height
-        ) < center[1] < (map_center[1] + height)
+        return (map_center[0] - width) < center[0] < (map_center[0] + width) and
```
Ruishenl, line 448:
```suggestion
        return (map_center[0] - width) < center[0] < (map_center[0] + width) and  \
```

Ruishenl (approved):

PR #201, opened 2024-04-09
Make pygame and cv2 dependency optional

Ruishenl:
- Use PIL instead of opencv to decode the encoded bvs to avoid import error when opencv is not installed
- Use Tuple instead of pygame's Vector2 in area_drive to make pygame also an optional dependency. This will pave the way for us to consider packaging the area_drive folder inside invertedai.

I was able to run the ` large_map_example.py` example in a minimum virtualenv with only `invertedai` installed and generating the bv.

@adscib  Let me know what do you think about this.

### DIFF
--- a/invertedai/common.py
+++ b/invertedai/common.py
@@ -2,7 +2,9 @@ from typing import List, Optional, Dict
 from enum import Enum
 from pydantic import BaseModel, model_validator
 import math
-
+from PIL import Image as PImage
+import numpy as np
+import io
 import invertedai as iai
 from invertedai.error import InvalidInputType, InvalidInput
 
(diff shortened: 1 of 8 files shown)


invertedai/common.py
```diff
@@ -2,7 +2,9 @@
 from enum import Enum
 from pydantic import BaseModel, model_validator
 import math
-
+from PIL import Image as PImage
+import numpy as np
```
adscib, line 6: According to the `pyproject.toml` file, `numpy` is not a required dependency, so we shouldn't be importing it here at top level. It's a little concerning that the CI passes despite this.

adscib (changes requested):
I generally like the change, as this is a lighter weight approach, but we still need to be careful about which dependencies we require.

invertedai/common.py
Ruishenl: I think it’s because matplotlib is a required dependency and numpy is actually a required package for it.

adscib: Why is matplotlib required in the first place? Is it just for ScenePlotter? In any case, I think we should make a conscious decision about whether we want to require numpy independently of matplotlib, and if so add it to the dependencies explicitly.

Ruishenl:
I added numpy explicitly as the required package and removed the both matplotlib and numpy from the dev dependencies

adscib (approved):
I trust that you checked that the birdviews still decode correctly.

PR #203, opened 2024-04-22
Large area simulation rework

KieranRatcliffeInvertedAI:
Moved versions of initialize and drive designed for simulating on large maps to a different submodule called "large" in invertedai. 

- Developed a new basic datatype called Region which contains a center, fov, and agent information such as state, attributes, and recurrent state as well as helper functions. This data type was designed with some thought towards the future in case basic assumptions about what a "region" can be changed (e.g. bigger or smaller than 100x100m square, non-square regions). 

- Due to the lack of need for the pygame visualization and design constraint of being stateless, region_drive has become simpler and less reliant on dependencies. 

- Region initialize is now separated from HOW the regions are produced. Default helper functions have been provided for this but can be easily swapped out by the user. 

- The core functionality is largely the same as it was before, algorithmically speaking. 

Questions for reviewers:
- What should we do concerning support for old versions of these utilities? 
- Do the comments make sense? Should more be added?
- Any other thoughts on the structure of the interface or future-proofing concerns?


invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+    AgentAttributes
+    """
+
+    center: Optional[Point] = None # The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
+    fov: Optional[float] = None # Side length of the region for the default interpretation of a region as a square
+    agent_states: Optional[List[AgentState]] = None # A list of existing agents within the region
+    agent_attributes: Optional[List[AgentAttributes]] = None # The attributes of agents that exist within the region or that will be initialized within the region
+    recurrent_states: Optional[List[RecurrentState]] = None # Recurrent states of the agents contained within the region
+    region_type: str = 'square' # Geometric category of the region. As of now, only 'square' is supported.
+    vertexes: List[Point] # An ordered list of x-y coordinates of the region defined clockwise around the perimeter of the region
```
Ruishenl, line 21: If you are referring to `vertex`, i believe the plural form is `vertices`.

invertedai/large/drive.py
```diff
@@ -0,0 +1,301 @@
...
+import asyncio
+
+import invertedai as iai
+from invertedai.large.common import Region
+from invertedai.common import Point, AgentState, AgentAttributes, RecurrentState, TrafficLightStatesDict, LightRecurrentStates
+from invertedai.api.drive import DriveResponse
+
+BUFFER_FOV = 35
+
+class Particle(BaseModel):
```
Ruishenl, line 13: Are we sure we want to keep using the word `Particle` here ? TreeNode or Node make more sense to me here. Or we can simply just call it Agent in this local namespace.

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+        vertexes = [
+            Point.fromlist([center_x-fov_split,center_y+fov_split]), # Top left
+            Point.fromlist([center_x+fov_split,center_y+fov_split]), # Top right
+            Point.fromlist([center_x+fov_split,center_y-fov_split]), # Bottom right
+            Point.fromlist([center_x-fov_split,center_y-fov_split]) # Bottom left
+        ]
+
+        return vertexes
+
+    def check_point_in_region(self,point):
```
Ruishenl, line 85: It makes sense to have a doc string for this method explaining how things work in general, but you should reformat it  to have it enclosed instead of using the comments. 
And i don't think you need those additional comments explain how things work line by line

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+                            # The point is on the edge between p1 and p2 therefore is within the region
+                            return True
+
+                        if x <= x_intersection:
+                            # Since the largest x value of the edge is to the right of the point and the intercept of the edge is 
+                            # also to the right, the ray crosses the edge. 
+                            is_inside = not is_inside
+            p1 = p2
+
+    def check_point_in_bounding_box(self,point):
```
Ruishenl, line 127: Same here. Need doc string instead

Ruishenl (commented):
Some initial comments

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+class Region(BaseModel):
+    """
+    A data structure to contain information about a region in a map to be used in large simulations.
+
+    See Also
+    --------
+    AgentAttributes
+    """
+
+    center: Optional[Point] = None # The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
```
adscib, line 15:
```suggestion
    center: Optional[Point] = None #: The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
```
Use `#:` for each field to include those comments in the documentation.

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+    fov: Optional[float] = None # Side length of the region for the default interpretation of a region as a square
```
adscib, line 16: `size` or `length` would be a better name, as the region definition doesn't necessarily have anything to do with viewing.

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
+from typing import List, Optional, Tuple
+from pydantic import BaseModel
+
+from invertedai.common import AgentAttributes, AgentState, RecurrentState, Point
+
+class Region(BaseModel):
```
adscib, line 6: If we only allow squares for the time being, it's better to remove all the currently unused fields. They serve no purpose currently and only commit you to some design choices that may turn out to be suboptimal in the future. You can always add them later. Similarly, making things optional in the future is a non-breaking change, so if the value for a given field is currently required, don't make it optional. I would remove everything except `centre` and the agent specifications.

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+    See Also
+    --------
+    AgentAttributes
+    """
+
+    center: Optional[Point] = None # The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
+    fov: Optional[float] = None # Side length of the region for the default interpretation of a region as a square
+    agent_states: Optional[List[AgentState]] = None # A list of existing agents within the region
+    agent_attributes: Optional[List[AgentAttributes]] = None # The attributes of agents that exist within the region or that will be initialized within the region
+    recurrent_states: Optional[List[RecurrentState]] = None # Recurrent states of the agents contained within the region
```
adscib, line 19: Is there a semantic difference between `None` and `[]` here? I'm also wondering if we should make it a `List[Tuple[AgentAttributes, AgentState, Optional[RecurrentState]]]` instead, but I'm not convinced.

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
...
+    region_type: str = 'square' # Geometric category of the region. As of now, only 'square' is supported.
+    vertexes: List[Point] # An ordered list of x-y coordinates of the region defined clockwise around the perimeter of the region
+
+    @classmethod
+    def fromlist(cls, l, agent_states = [], agent_attributes = [], recurrent_states = [], region_type = 'square'):
```
adscib, line 24: This method is a nightmare for backwards compatibility. Do we actually need it? I don't think we're sending those objects through the API.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+from invertedai.error import InvertedAIError
+
+SLACK = 2
+AGENT_SCOPE_FOV_BUFFER = 20
+
+@validate_call
+def define_regions_grid(
+    map_center: Optional[Tuple[float,float]] = (0.0,0.0),
+    width: Optional[float] = 100.0, 
+    height: Optional[float] = 100.0,
```
adscib, line 22: Why are these optional? I can't think of a reason you would want to grid an area without knowing its size or position.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+from invertedai.api.initialize import InitializeResponse
+from invertedai.common import TrafficLightStatesDict, Point
+from invertedai.utils import get_default_agent_attributes
+from invertedai.error import InvertedAIError
+
+SLACK = 2
+AGENT_SCOPE_FOV_BUFFER = 20
+
+@validate_call
+def define_regions_grid(
```
adscib, line 19: Needs a docstring explaining how the grid is produced and what is a typical use case.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def get_neighbors(center):
+        return [
+            (center[0] + (i * stride), center[1] + (j * stride))
+            for i, j in list(product(*[(-1, 1),]* 2))
+        ]
+
+    queue, centers = [map_center], []
+
+    while queue:
```
adscib, line 37: BFS is way of an overkill here. You can compute the grid of a rectangular area in closed form.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+        regions[i] = Region.fromlist([Point.fromlist(list(center))])
+
+    return regions
+
+@validate_call
+def get_regions_density_by_road_area(
+    location: str,
+    regions: List[Region],
+    max_agent_density: Optional[int] = 10,
+    scaling_factor: Optional[float] = 1.0,
```
adscib, line 59: These argument names, as well as the name of the function itself, are not at all intuitive. Presumably what this function does is compute agent numbers for each region given some parameters, so the naming should reflect that. Neither `density` nor `scaling` have obvious meanings here.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+    display_progress_bar: Optional[bool] = True
+) -> List[Region]:
+    #Get fraction of image that is a drivable surface (assume all non-black pixels are drivable)
+    center_road_area_dict = {}
+    max_drivable_area_ratio = 0
+    
+    iterable_regions = None
```
adscib, line 66:
```suggestion
```

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+            rendering_center=center_tuple
+        ).birdview_image.decode()
+
+        ## Get number of black pixels
+        birdview_arr_shape = birdview.shape
+        total_num_pixels = birdview_arr_shape[0]*birdview_arr_shape[1]
+        # Convert to grayscale using Luminosity Method: gray = 0.114*B + 0.587*G + 0.299*R
+        # Image should be in BGR color pixel format
+        birdview_grayscale = np.matmul(birdview.reshape(total_num_pixels,3),np.array([[0.114],[0.587],[0.299]]))
+        number_of_black_pix = np.sum(birdview_grayscale == 0)
```
adscib, line 91:
```suggestion
        number_of_black_pix = np.sum((birdview.sum(axis=-1) == 0).int())
```
This is a very elaborate conversion only to compare things to zero at the end.

adscib (changes requested):
Here's some feedback for the parts I managed to read tonight. I'll add more tomorrow. Generally, the structure is looking good and this is what we want, but there are a lot of things that can be simplified and renamed to something more obvious.

If we're confident that the old versions of those functions are not used, I think it's fine to delete them. If we thought they might be used, deprecating them for some time would be in order.

One name change to consider is to use `large_initialize` and `large_drive` instead of `region_initialize` and `region_drive`. Those are more in line with what those functions accomplish rather than how they do it.

invertedai/large/common.py
KieranRatcliffeInvertedAI: Vertices is indeed used but Vertexes is also used and more grammatically correct.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: This was a carry over from the old area drive, however particle is the most commonly used term for "things that are placed within a quad tree". However Agent would indeed be more effective and descriptive.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Hm yeah I guess at this point we know we don't need to do anything more fancy with a grayscale image. I made this change.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Deleted.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Good point, made these parameters mandatory.

invertedai/large/common.py
KieranRatcliffeInvertedAI: If these Regions were only used for driving then I would agree with the Tuple but since it's used for initialize as well which can be quite wonky, I'd say lets keep state, attributes, and recurrent states separate.

invertedai/large/drive.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def get_leaf_nodes(self):
+        if self.leaf:
+            return [self]
+        else:
+            return self.northWest.get_leaf_nodes() + self.northEast.get_leaf_nodes() + \
+                self.southWest.get_leaf_nodes() + self.southEast.get_leaf_nodes()
+
+def _flatten_and_sort(nested_list,index_list):
+    # Cannot unpack iterables during list comprehension so use this helper function instead
```
rf-ivtdai, line 132: remove?

invertedai/large/drive.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+    
+    num_agents = len(agent_states)
+    agent_x = [None]*num_agents
+    agent_y = [None]*num_agents
+    for i, agent in enumerate(agent_states):
+        agent_x[i] = agent.center.x
+        agent_y[i] = agent.center.y
+    max_x, min_x, max_y, min_y = max(agent_x), min(agent_x), max(agent_y), min(agent_y)
+    region_fov = ceil(max(max_x,max_y) - min(min_x,min_y)) #Round up the starting FOV to an integer to reduce floating point issues
```
rf-ivtdai, line 241: Consider removing the comment

invertedai/large/drive.py
```diff
@@ -0,0 +1,301 @@
...
+    agent_id_order = []
+    for i, leaf_node in enumerate(all_leaf_nodes):
+        region, region_buffer = leaf_node.region, leaf_node.region_buffer
+        region_agents_ids = [particle.agent_id for particle in leaf_node.particles]
+        
+        if len(region.agent_states) == 0:
+            # Region is empty, do not call DRIVE
+            continue
+        
+        else:
```
rf-ivtdai, line 265:
```suggestion
        if len(region.agent_states) > 0:
```

invertedai/large/common.py
```diff
@@ -0,0 +1,153 @@
+from typing import List, Optional, Tuple
+from pydantic import BaseModel
+
+from invertedai.common import AgentAttributes, AgentState, RecurrentState, Point
+
+class Region(BaseModel):
+    """
+    A data structure to contain information about a region in a map to be used in large simulations.
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
KieranRatcliffeInvertedAI: I've kept size (formerly fov) for now since that is necessary for region_drive but otherwise everything else has been removed.

invertedai/large/common.py
KieranRatcliffeInvertedAI: I replaced it with "init_square_region()" since the assert check is useful in some contexts and having this as an entry point might be useful if we want to change how the object is initialized.

rf-ivtdai (commented):
Generally I see 3 types of comments:

1. definition to arguments and fields
2. explanation of key steps within a long workflow
3. explanation of why things are called/implemented as such

For this repo, I believe type 1 and 2 are reasonable, but we should remove/rephrase those that are type 3. I marked them below.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: I've renamed them but when could you provide some feedback on if the new names are insufficient, too wordy, and if so, could you elaborate on what you are looking for?

invertedai/large/common.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: For the sake of time, I can update this later since this won't impact the user interface and we know this code works so we can quietly make that change later while users are testing the features?

invertedai/large/drive.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+                random_seed=random_seed,
+                get_birdview=get_birdview,
+            )
+
+            if traffic_light_state_history is None and response.traffic_lights_states is not None:
+                traffic_light_state_history = [response.traffic_lights_states]
+
+        except InvertedAIError as e:
+            iai.logger.warning(e)
+            continue
```
adscib, line 266: It's a little odd to just ignore this. This breaks the usual contract of INITIALIZE, which is that you get exactly as many agents as you request, or it's a failure. Let's discuss in person.

invertedai/large/initialize.py
adscib: Yes, it can be done later.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+            desc=f"Initializing regions"
+        )
+    else:
+        iterable_regions = regions
+
+    for i, region in iterable_regions:
+        region_center = region.center
+        region_size = region.size
+
+        existing_agent_states, existing_agent_attributes, _ = _get_all_existing_agents_from_regions(regions)
```
adscib, line 235: I think you can pull this out of the loop?

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+
+        region_predefined_agent_states = [] if region.agent_states is None else region.agent_states
+        all_agent_states = conditional_agent_states + region_predefined_agent_states
+        all_agent_attributes = conditional_agent_attributes + region.agent_attributes
+
+        try:
+            response = iai.initialize(
+                location=location,
+                states_history=None if len(all_agent_states) == 0 else [all_agent_states],
+                agent_attributes=[] if len(all_agent_attributes) == 0 else all_agent_attributes,
```
adscib, line 253:
```suggestion
                agent_attributes=all_agent_attributes,
```

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+        conditional_agents = list(filter(
+            lambda x: inside_fov(center=region_center, agent_scope_fov=region_size+AGENT_SCOPE_FOV_BUFFER, point=x[0].center), 
+            zip(existing_agent_states,existing_agent_attributes)
+        ))
+
+        conditional_agent_states = [x[0] for x in conditional_agents]
+        conditional_agent_attributes = [x[1] for x in conditional_agents]
+
+        region_predefined_agent_states = [] if region.agent_states is None else region.agent_states
+        all_agent_states = conditional_agent_states + region_predefined_agent_states
```
adscib, line 246: Aren't those already captured in `conditional_agent_states`? It looks like you're including them twice.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+        response_agent_attributes_sampled = response.agent_attributes[len(all_agent_states):]
+        response_recurrent_states_sampled = response.recurrent_states[len(all_agent_states):]
+
+        # Filter out agents that are not inside the ROI to avoid collision with other agents not passed as conditional
+        # SLACK is for removing the agents that are very close to the boundary and they may collide with agents not 
+        # labelled as conditional
+        valid_agents = list(filter(
+            lambda x: inside_fov(center=region_center, agent_scope_fov=region_size - SLACK, point=x[0].center),
+            zip(response_agent_states_sampled, response_agent_attributes_sampled, response_recurrent_states_sampled)
+        ))
```
adscib, line 279: This is another instance where we remove some agents that the user is expecting to see. Perhaps this is fine, but we need to be clear in the documentation about what we return.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+
+        valid_agent_state = [x[0] for x in valid_agents]
+        valid_agent_attrs = [x[1] for x in valid_agents]
+        valid_agent_rs = [x[2] for x in valid_agents]
+
+        predefined_agents_slice = slice(len(conditional_agent_states),len(conditional_agent_states)+len(region_predefined_agent_states))
+        regions[i].agent_states = response.agent_states[predefined_agents_slice] + valid_agent_state
+        regions[i].agent_attributes = response.agent_attributes[predefined_agents_slice] + valid_agent_attrs
+        regions[i].recurrent_states = response.recurrent_states[predefined_agents_slice] + valid_agent_rs
```
adscib, line 288: Modifying Region objects in place is another thing that we need to be careful about in the documentation.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+        valid_agents = list(filter(
+            lambda x: inside_fov(center=region_center, agent_scope_fov=region_size - SLACK, point=x[0].center),
+            zip(response_agent_states_sampled, response_agent_attributes_sampled, response_recurrent_states_sampled)
+        ))
+
+        valid_agent_state = [x[0] for x in valid_agents]
+        valid_agent_attrs = [x[1] for x in valid_agents]
+        valid_agent_rs = [x[2] for x in valid_agents]
+
+        predefined_agents_slice = slice(len(conditional_agent_states),len(conditional_agent_states)+len(region_predefined_agent_states))
```
adscib, line 285: Are you sure this is correct? I don't know what exactly was the intention here, but this seems suspicious to me.

adscib (changes requested):
Here's a more careful review of `initialize`. I'll review `drive` as well, but posting this already in the interest of expediency.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Done

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: The regions are populated with agents one at a time. Therefore, each loop all agents currently in the simulation (initialized, preddefined, whatever) need to be accessed so that ones that are relevant to the current region of interest can be included as conditional. An alternative would be to have 3 running lists for states, attributes, and recurrent states but I went with this method for the sake of saving memory space and code cleanliness at the cost of a small increase in computation time due to looping through every region.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+            # Initialize simulation with an API call
+            response = iai.initialize(
+                location=location,
+                states_history=None if len(all_agent_states) == 0 else [all_agent_states],
+                agent_attributes=[] if len(all_agent_attributes) == 0 else all_agent_attributes,
+                get_infractions=get_infractions,
+                traffic_light_state_history=traffic_light_state_history,
+                location_of_interest=(region_center.x, region_center.y),
+                random_seed=random_seed,
+                get_birdview=get_birdview,
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: I'm fine with leaving this out for now because it's not-trivial how to interpret this in the "large" version of initialize and drive and there are indeed other visualization methods.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,265 @@
...
+
+        # if get_birdview:
+        #     file_path = f"{save_birdviews_to}_{(region_center.x, region_center.y)}.jpg"
+        #     response.birdview.decode_and_save(file_path)
+
+    all_agent_states, all_agent_attributes, all_recurrent_states = _get_all_existing_agents_from_regions(regions)
+
+    response.agent_states = all_agent_states
+    response.agent_attributes = all_agent_attributes
+    response.recurrent_states = all_recurrent_states
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: I see. I've made a change that should account for this.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: The difference here is that `conditional_agent_states` contains states from regions that have already been initialized and predefined agents from other regions whereas `region_predefined_agent_states` contains predefined agents for the current region of interest. Any of these agents could be relevant as "conditional" for calling `initialize` on the current region.

KieranRatcliffeInvertedAI: I've renamed `conditional_agent_states` to `out_of_region_conditional_agent_states` and used this naming scheme wherever else relevant.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Actually upon further inspection, this can be removed because of filter earlier in the loop that only takes in agents as conditional if they are within a certain scope so no collisions should happen at the edge anyways.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: However, now with the deletion of that "valid states" filter, this is moot and the slice can just be `len(out_of_region_conditional_agent_states)` onwards.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Fixed so that it uses the member function `region[i].insert_all_agent_details` instead of modifying the fields directly.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+            }
+            if not is_async:
+                all_responses.append(iai.drive(**input_params))
+            else:
+                async_input_params.append(input_params)
+
+    if is_async:
+        all_responses = asyncio.run(async_drive_all(async_input_params))
+
+    response = DriveResponse(
```
adscib, line 303: This whole `_flatten_and_sort` business is quite convoluted and could use refactoring, but we can leave that for later.

adscib (changes requested):
Here's a review for `drive`. I'll do one more read of the whole thing.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    recurrent_states: Optional[List[RecurrentState]] = None,
+    traffic_lights_states: Optional[TrafficLightStatesDict] = None,
+    light_recurrent_states: Optional[LightRecurrentStates] = None,
+    get_birdview: bool = False,
+    rendering_center: Optional[Tuple[float, float]] = None,
+    rendering_fov: Optional[float] = None,
+    get_infractions: bool = False,
+    random_seed: Optional[int] = None,
+    api_model_version: Optional[str] = None,
+    capacity: Optional[int] = 100,
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: This makes sense, change is implemented.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    is_async: Optional[bool] = True
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+
+    is_async:
+        A flag to control whether to use asynchronous DRIVE calls.
+
+    See Also
+    --------
+    :func:`drive`
+    """
+    
+    num_agents = len(agent_states)
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    agent_x = [None]*num_agents
+    agent_y = [None]*num_agents
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    for i, agent in enumerate(agent_states):
+        agent_x[i] = agent.center.x
+        agent_y[i] = agent.center.y
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    max_x, min_x, max_y, min_y = max(agent_x), min(agent_x), max(agent_y), min(agent_y)
+    region_size = ceil(max(max_x,max_y) - min(min_x,min_y))
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: The reason this is rounded up to the nearest integer is to avoid issues with floating points as the region size is divided over and over. In testing, this would create a small, but non-zero gap, between leaf nodes in the quad tree and this was the best option for solving that issue. 

Otherwise the change is good and implemented.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    region_center = ((max_x+min_x)/2,(max_y+min_y)/2)
+
+    quadtree = QuadTree(
+        capacity=capacity,
+        region=Region.init_square_region(
+            Point.fromlist(list(region_center)),
+            region_size
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+        ),
+    )
+    for i, (agent, attrs, recurr_state) in enumerate(zip(agent_states,agent_attributes,recurrent_states)):
+        agent_info = AgentInfo.fromlist([agent, attrs, recurr_state, i])
+        is_inserted = quadtree.insert(agent_info)
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: It would only ever be false if the agent was outside of the regions defined by the quad tree which would be an error so I've added an appropriate error message.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+import asyncio
+
+import invertedai as iai
+from invertedai.large.common import Region
+from invertedai.common import Point, AgentState, AgentAttributes, RecurrentState, TrafficLightStatesDict, LightRecurrentStates
+from invertedai.api.drive import DriveResponse
+
+BUFFER_FOV = 35
+
+class AgentInfo(BaseModel):
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+        recurrent_states = _flatten_and_sort([response.recurrent_states[:len(region.agent_attributes)] for response, region in zip(all_responses,non_empty_regions)],agent_id_order),
+        is_inside_supported_area = _flatten_and_sort([response.is_inside_supported_area[:len(region.agent_attributes)] for response, region in zip(all_responses,non_empty_regions)],agent_id_order),
+        infractions = [] if not get_infractions else _flatten_and_sort([response.infractions[:len(region.agent_attributes)] for response, region in zip(all_responses,non_empty_regions)],agent_id_order),
+        api_model_version = '' if len(all_responses) == 0 else all_responses[0].api_model_version,
+        birdview = None if len(all_responses) == 0 or not get_birdview else all_responses[0].birdview,
+        traffic_lights_states = traffic_lights_states if len(all_responses) == 0 else all_responses[0].traffic_lights_states,
+        light_recurrent_states = light_recurrent_states if len(all_responses) == 0 else all_responses[0].light_recurrent_states
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Good point, I've made that change as well as adding an if statement to check whether it should just call regular drive (i.e. if there is only 1 leaf node).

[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: If the user happens to input no agent information, the regular call to drive() will handle that.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+        if self.leaf:
+            return [self]
+        else:
+            return self.northWest.get_leaf_nodes() + self.northEast.get_leaf_nodes() + \
+                self.southWest.get_leaf_nodes() + self.southEast.get_leaf_nodes()
+
+def _flatten_and_sort(nested_list,index_list):
+    flat_list = []
+    for sublist in nested_list:
+        flat_list.extend(sublist)
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+    
+    sorted_list = []
+    for ind in range(len(index_list)):
+        agent_index = index_list.index(ind)
+        sorted_list.append(flat_list[agent_index])
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: I'm not sure why the zip is there but I think using the index_list as the key argument in sorted() is the way to go. I.e.:
`sorted_list = sorted(flat_list,index_list)`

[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def tolist(self):
+        return [self.agent_state, self.agent_attributes, self.recurrent_state, self.agent_id]
+
+    @classmethod
+    def fromlist(cls, l):
+        agent_state, agent_attributes, recurrent_state, agent_id = l
+        return cls(agent_state=agent_state, agent_attributes=agent_attributes, recurrent_state=recurrent_state, agent_id=agent_id)
+
+class QuadTree:
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

KieranRatcliffeInvertedAI:
Outstanding issues:

1. Despite some small buffer room from rounding up the region_size for the root layer of the quadtree, some gaps still exist between leaf nodes due to floating point issues when dividing the size of the leaf nodes over and over.
2. I've only seen this once but I got a recursion limit error, perhaps because there was one region on the map that was too dense and and, because the neighbouring regions were also dense, could not subdivide to a size of leaf node that contained less than the capacity.

invertedai/large/drive.py
```diff
@@ -0,0 +1,314 @@
...
+            if not is_async:
+                all_responses.append(iai.drive(**input_params))
+            else:
+                async_input_params.append(input_params)
+
+    if is_async:
+        all_responses = asyncio.run(async_drive_all(async_input_params))
+
+    response = DriveResponse(
+        agent_states = _flatten_and_sort([response.agent_states[:len(region.agent_attributes)] for response, region in zip(all_responses,non_empty_regions)],agent_id_order),
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Actually scratch that, if you attempt to refactor the zip part into a single variable, I think there is some pass-by-reference shenanigans where that zipped variable is being affected by the `_flatten_and_sort` function and that's why I had to repeat it before.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: It is an annoying circumstance for sure. There is no connection between the location of the agent and it's index in the input parameter. This is a situation where an explicit ID for each agent would be useful but I understand it wouldn't make sense to change one of our basic data classes just to solve this. 

I think a middle ground is that this flatten and sort method can be a member function of the quadtree class instead to make it a bit cleaner remaining functionally the same.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+@validate_call
+def get_regions_in_grid(
+    width: float,
+    height: float,
+    map_center: Optional[Tuple[float,float]] = (0.0,0.0), 
+    stride: Optional[float] = 50.0
+) -> List[Region]:
+    """
+    A utility function to help initialize an area larger than 100x100m. This function breaks up an 
+    area into a grid of 100x100m regions, each of which can be initialized directly.
```
adscib, line 96: Clarify whether the returned regions have any agents.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+def get_number_of_agents_per_region_by_drivable_area(
+    location: str,
+    regions: List[Region],
+    total_num_agents: Optional[int] = 10,
+    random_seed: Optional[int] = None,
+    display_progress_bar: Optional[bool] = True
+) -> List[Region]:
+    """
+    This function takes in a list of regions, infers the total drivable surface area per region by calculating 
+    the number of black pixels (non-drivable surfaces), then inserts a number of default AgentAttributes objects
```
adscib, line 152:
```suggestion
    This function takes in a list of regions, calculates the driveable area for each of them
    using :func:`location_info`, then inserts a number of default AgentAttributes objects
```
The implementation details are just confusing, but we want to let the user know which API endpoints will be called.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+    return regions
+
+@validate_call
+def get_number_of_agents_per_region_by_drivable_area(
+    location: str,
+    regions: List[Region],
+    total_num_agents: Optional[int] = 10,
+    random_seed: Optional[int] = None,
+    display_progress_bar: Optional[bool] = True
+) -> List[Region]:
```
adscib, line 149: You're returning a list of regions. Make sure to specify whether those are the input Region objects, and if not whether the input objects are modified.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+            assert cls.check_point_in_bounding_box(cls,agent.center), f"Existing agent states at position {agent.center} must be located within the region."
+
+        return cls(
+            center=center, 
+            size=size,
+            agent_states=agent_states, 
+            agent_attributes=agent_attributes, 
+            recurrent_states=recurrent_states 
+        )
+
```
adscib, line 40: A `copy` method would be a good idea for future proofing.

invertedai/large/drive.py
```diff
@@ -0,0 +1,188 @@
...
+    random_seed: Optional[int] = None,
+    api_model_version: Optional[str] = None,
+    single_call_agent_limit: Optional[int] = None,
+    async_api_calls: bool = True
+) -> DriveResponse:
+    """
+    A utility function to drive more than the normal capacity of agents in a call to :func:`drive`.
+    The agents are inserted into a quadtree structure and :func:`drive` is then called on each
+    region represented by a leaf node of the quadtree. Agents near this region are included in the
+    :func:`drive` calls to ensure the agents behave properly locally. The quadtree is constructed
```
adscib, line 34:
```suggestion
    :func:`drive` calls to ensure the agents see all their neighbours. The quadtree is constructed 
```

invertedai/large/drive.py
```diff
@@ -0,0 +1,188 @@
...
+       If the location contains traffic lights within the supported area,
+       their current state should be provided here. Any traffic light for which no
+       state is provided will have a state generated by iai.
+
+    light_recurrent_states:
+       Light recurrent states for all agents, obtained from the previous call to
+        :func:`drive` or :func:`initialize` or the large equivalents.
+       Specifies the state and time remaining for each light group in the map.
+       If manual control of individual traffic lights is desired, modify the relevant state(s) 
+       in traffic_lights_states, then pass in light_recurrent_states as usual.
```
adscib, line 69: Consider pointing the reader to the documentation for drive instead. That has its downsides, but makes it less likely that this description gets out of sync from the one in drive.

invertedai/large/drive.py
```diff
@@ -0,0 +1,188 @@
...
+        Controls the stochastic aspects of agent behavior for reproducibility.
+
+    api_model_version:
+        Optionally specify the version of the model. If None is passed which is by default, 
+        the best model will be used.
+    
+    single_call_agent_limit:
+        The number of agents allowed in a region before it must subdivide. Currently this 
+        value represents the capacity of a quadtree leaf node that will subdivide if the 
+        number of vehicles in the region passes this threshold.
```
adscib, line 85:
```suggestion
        The number of agents allowed in a region before it must subdivide. Currently this 
        value represents the capacity of a quadtree leaf node that will subdivide if the 
        number of vehicles in the region passes this threshold. In any case,
        this will be limited to the maximum currently supported by :func:`drive`.
```
Assuming we clip this value at 100, which I think we should do.

adscib (changes requested):
Nice, it looks like we have the major things covered. Other than for some final tidying up, we should be good to go. Switching to an iterative QuadTree implementation would be my recommendation for getting rid of the recursion depth errors, but that can wait for later. More importantly, we need some test to ensure those new functions won't get broken by subsequent commits, but that can also be done in a separate PR.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    A region in a map used to divide a large simulation into smaller parts.
+
+    See Also
+    --------
+    AgentAttributes
+    """
+
+    center: Optional[Point] = None #: The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
+    size: Optional[float] = None #: Side length of the region for the default interpretation of a region as a square
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Ah yes a oversight when this was changed from the more general Region, this has been updated.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+    agent_states: Optional[List[AgentState]] = [] #: A list of existing agents within the region
+    agent_attributes: Optional[List[AgentAttributes]] = [] #: The attributes of agents that exist within the region or that will be initialized within the region
+    recurrent_states: Optional[List[RecurrentState]] = [] #: Recurrent states of the agents contained within the region
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Correct, if recurrent_states is not empty, it should be the same length as agent_states AND agent_attributes. The two uses cases that I can identify for this object are for passing as arguments to DRIVE and to INITIALIZE which have slightly different needs so it's hard to enforce a standard across the board, especially because INITIALIZE has many different operating modes (i.e. equal number states and attributes, more attributes than states, no states at all).

Do you think it'd be worthwhile to split these regions up? Or have some sort of data validation? E.g. if agent details are to be added, it should check that it is a valid configuration?

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    def init_square_region(
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+            agent_states=agent_states, 
+            agent_attributes=agent_attributes, 
+            recurrent_states=recurrent_states 
+        )
+
+    def clear_agents(self):
+        """
+        Helper function to clear all agent data within the Region but keep geometric information.
+
+        """
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self.agent_states = []
+        self.agent_attributes = []
+        self.recurrent_states = []
+
+    def insert_all_agent_details(self,agent_state,agent_attributes,recurrent_state):
+        """
+        Helper function to insert the details relevant to a single agent.
+
+        """
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self.agent_states.append(agent_state)
+        self.agent_attributes.append(agent_attributes)
+        self.recurrent_states.append(recurrent_state)
+
+    def check_point_in_bounding_box(self,point):
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
...
+        region is rectangular.
+
+        """
+
+        x, y = point.x, point.y
+        region_x, region_y = self.center.x, self.center.y
+        if region_x - self.size/2 <= x and x <= region_x + self.size/2 and region_y - self.size/2 <= y and y <= region_y + self.size/2:
+            return True
+        else:
+            return False
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
```diff
@@ -0,0 +1,74 @@
+from typing import List, Optional, Tuple
+from pydantic import BaseModel
+
+from invertedai.common import AgentAttributes, AgentState, RecurrentState, Point
+
+class Region(BaseModel):
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+        Controls the stochastic aspects of initialization for reproducibility.
+
+    api_model_version:
+        Optionally specify the version of the model. If None is passed which is by default, the best model will be used.
+
+    display_progress_bar:
+        If True, a bar is displayed showing the progress of all relevant processes.
+
+    return_exact_agents:
+        If set to True, this function will raise an Exception if it cannot fit the requested number of agents in any single
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+    if :func:`initialize` is unable to produce viable vehicle placements if the initialization
+    should continue or raise an exception.
+
+    Arguments
+    ----------
+    location:
+        Location name in IAI format.
+
+    regions:
+        List of regions that contains information about the location, shape, and number of agents within the region.
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done and added a note to refer to the initialize documentation for details it covers there.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+    api_model_version: Optional[str] = None,
+    display_progress_bar: bool = True,
+    return_exact_agents: bool = False
+) -> InitializeResponse:
+    """
+    A utility function to initialize an area larger than 100x100m. This function breaks up an 
+    area into a grid of 100x100m regions and runs initialize on them all. For any particular 
+    region of interest, existing agents near enough to the region of interest are passed as 
+    conditional agents to :func:`initialize`. Regions will be rejected and :func:`initialize` 
+    will not be called if it is not possible for agents to exist there (e.g. there are no
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: If you're referring to the "Regions will be rejected..." I changed this to be more accurate (i.e. initialize won't be called if there are no agents requested.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+    """
+    A utility function to initialize an area larger than 100x100m. This function breaks up an 
+    area into a grid of 100x100m regions and runs initialize on them all. For any particular 
+    region of interest, existing agents near enough to the region of interest are passed as 
+    conditional agents to :func:`initialize`. Regions will be rejected and :func:`initialize` 
+    will not be called if it is not possible for agents to exist there (e.g. there are no 
+    drivable surfaces present). As well, predefined agents may be passed via the regions and 
+    will be considered as conditional. A boolean flag can be used to control failure behaviour
+    if :func:`initialize` is unable to produce viable vehicle placements if the initialization
+    should continue or raise an exception.
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+                        agent_attributes=all_agent_attributes,
+                        get_infractions=get_infractions,
+                        traffic_light_state_history=traffic_light_state_history,
+                        location_of_interest=(region_center.x, region_center.y),
+                        random_seed=random_seed
+                    )
+
+                except InvertedAIError as e:
+                    # If error has occurred, display the warning and retry
+                    iai.logger.warning(f"Region initialize attempt {attempt} error: {e}")
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+        height, width = 100, 100
+        polygon_x, polygon_y = [], []
+        for pt in location_response.bounding_polygon:
+            polygon_x.append(pt.x)
+            polygon_y.append(pt.y)
+        width = max(polygon_x) - min(polygon_x)
+        height = max(polygon_y) - min(polygon_y)
+
+        area_size = tuple([width,height])
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done. I also changed this variable name to area_shape to be slightly more accurate and future proof as well as be terminologically proximal to other python packages like numpy.

invertedai/large/initialize.py
```diff
@@ -0,0 +1,391 @@
...
+        Location name in IAI format.
+
+    total_num_agents:
+        The total number of agents to initialize across all regions.
+
+    area_size:
+        Contains the [width, height] of the rectangular area to be broken into smaller regions.
+
+    map_center:
+        The coordinates of the center of the rectangular area to be broken into smaller regions.
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/common.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: Done.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: I like this idea, I will do the same with `large_initialize` wherever possible because yeah it might be annoying but it is much more convenient for maintainability and less prone to mistakes and confusion.

invertedai/large/drive.py
KieranRatcliffeInvertedAI: Done.

KieranRatcliffeInvertedAI:
Outstanding addition to include:
- Tests for large initialize
- Tests for large drive

invertedai/large/common.py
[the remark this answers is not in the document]
adscib: Yes, I would perform some validation.

adscib (approved):

Ruishenl:
Before we merge, please make sure all the tests have passed. Right now we have been getting this error from the log:
```
tests/test_large_drive.py:38: in <module>
    initial_response = large_initialize(
../../../.cache/pypoetry/virtualenvs/invertedai-mfDNwQNM-py3.10/lib/python3.10/site-packages/pydantic/_internal/_validate_call.py:113: in __call__
    res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
invertedai/large/initialize.py:368: in large_initialize
    raise InvertedAIError(message=exception_string)
E   invertedai.error.InvertedAIError: Unable to initialize region at x=79.8 y=-4.3 with size 100.0 after 1 attempts. 
```

KieranRatcliffeInvertedAI:
> Before we merge, please make sure all the tests have passed. Right now we have been getting this error from the log:
> 
> ```
> tests/test_large_drive.py:38: in <module>
>     initial_response = large_initialize(
> ../../../.cache/pypoetry/virtualenvs/invertedai-mfDNwQNM-py3.10/lib/python3.10/site-packages/pydantic/_internal/_validate_call.py:113: in __call__
>     res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
> invertedai/large/initialize.py:368: in large_initialize
>     raise InvertedAIError(message=exception_string)
> E   invertedai.error.InvertedAIError: Unable to initialize region at x=79.8 y=-4.3 with size 100.0 after 1 attempts. 
> ```

Absolutely. I ran those tests locally and they passed without issue so I think therein lies the cause for the failure. I will come back and solve this issue before merging.

PR #205, opened 2024-04-24
Traffic light doc string updates

rf-ivtdai:
Update `traffic_lights_states` and `light_recurrent_states` doc strings in initialize and drive request and response.


adscib (commented):
Make sure to also update `docs/source/userguide.md`.

rf-ivtdai:
Screenshot from 2024-04-24 16-14-07

Built html docs page locally, can see `userguide.md` is updated,  `model_config` and `model_fields` are excluded from pythonapi docs. Doc strings in request and response fields needs to be merged to the `documentation` branch before it can take effect in the docs

invertedai/api/drive.py
```diff
@@ -102,14 +102,16 @@ def drive(
...
+       the original light states given by 'light_recurrent_states'.
 
     light_recurrent_states:
        Light recurrent states for all agents, obtained from the previous call to
         :func:`drive` or :func:`initialize`.
        Specifies the state and time remaining for each light group in the map.
-       If manual control of individual traffic lights is desired, modify the relevant state(s) 
+       To let iai manage all light states in the scene, simply pass 'light_recurrent_states' from the previous response of :func:`drive` here.
+       If customized control of individual traffic lights is desired, modify the relevant state(s) 
        in traffic_lights_states, then pass in light_recurrent_states as usual.
```
adscib, line 115: Does this actually work? I thought that once you pass the light states, it's game over for the server controller.

adscib (changes requested):
Thanks. We just need some clarification regarding whether the user can continue relying on the recurrent light states once they provide specific light states at some step. I was under the impression that they can't but maybe that's changed.

docs/source/userguide.md
```diff
@@ -64,7 +64,8 @@ of the map as well. Traffic light state changes dynamically and is controlled ex
 API. Each traffic light can be green, yellow, or red at any given point. Traffic light IDs are fixed and can be derived
 from the map, but for convenience we also provide traffic light IDs and the corresponding locations in LOCATION_INFO.
 For maps with traffic lights, the client is responsible for specifying their state on each call to INITIALIZE and DRIVE.
-If no state is provided for any particular light, it will be considered absent.
+This can be done either with manual setting of each light's state through 'traffic_lights_states', or let iai manage 
+light states through 'light_recurrent_states'.
```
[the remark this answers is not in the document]
rf-ivtdai: Updated all doc strings according to the comments. I think things are pretty clear in the request and response, but when putting everything together here, I'm not sure if this is both clear and succinct for the purpose of a general user guide.

invertedai/api/drive.py
```diff
@@ -102,15 +102,19 @@ def drive(
 
     traffic_lights_states:
        If the location contains traffic lights within the supported area,
-       their current state should be provided here. Any traffic light for which no
-       state is provided will have a state generated by iai.
+       their current state can be provided here. It is legal to not provide this field, and use
+       'light_recurrent_states' to step the traffic lights. If provided, light states from 'traffic_light_states' will override
+       the original light states given by 'light_recurrent_states'. The server does not currently support continuing user-provided light state sequences, 
+       so once the states are provided at any step, they should also be provided on all subsequent steps to guarantee coherent light sequences.
+       If neither 'traffic_lights_states' nor 'light_recurrent_states' are provided, the server will realistically initialize the traffic light states,
```
adscib, line 109:
```suggestion
       If neither 'traffic_lights_states' nor 'light_recurrent_states' are provided, the server will arbitrarily initialize the traffic light states,
```
I don't think the server will consider the current vehicle positions when setting the light states, so I wouldn't call it realistic.

adscib (approved):
Nice. I think it will have to do for the user guide for now. Just one nitpick left.

PR #207, opened 2024-04-30
Adding AgentProperties field for DRIVE and INITIALIZE

Ruishenl:


Ruishenl:
Both cpp and python updated

rf-ivtdai (approved):

PR #213, opened 2024-07-17
Updated all large area tools to use agent properties by default while…

KieranRatcliffeInvertedAI:
… maintaining backwards compatibility. 

This will be an official fix for the large simulation stack that was temporarily fixed before in an unmerged branch. This has been tested by passing both `AgentProperties` and `AgentAttributes` into the `large_map_example.py` script into both the `Regions` and the `large_drive` function as the two main access points where it could be either data type. 

If you find or think of anywhere else this change should be made (except other examples since that will be another branch), please let me know.

### DIFF
--- a/invertedai/large/common.py
+++ b/invertedai/large/common.py
@@ -1,7 +1,7 @@
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 from pydantic import BaseModel
 
-from invertedai.common import AgentAttributes, AgentState, RecurrentState, Point
+from invertedai.common import AgentAttributes, AgentProperties, AgentState, RecurrentState, Point
 
 class Region(BaseModel):
     """
@@ -9,13 +9,13 @@ class Region(BaseModel):
 
     See Also
     --------
-    AgentAttributes
+    AgentProperties
     """
 
     center: Point #: The center of the region if such a concept is relevant (e.g. center of a square, center of a rectangle)
     size: float #: Side length of the region for the default interpretation of a region as a square
     agent_states: Optional[List[AgentState]] = [] #: A list of existing agents within the region
-    agent_attributes: Optional[List[AgentAttributes]] = [] #: The attributes of agents that exist within the region or that will be initialized within the region
+    agent_properties: Optional[List[AgentProperties]] = [] #: The static parameters of agents that exist within the region or that will be initialized within the region
     recurrent_states: Optional[List[RecurrentState]] = [] #: Recurrent states of the agents eixsting within the region
 
     @classmethod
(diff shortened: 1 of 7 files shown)


adscib (commented):
I don't see anything suspicious here, but I'll let the others review it.

KieranRatcliffeInvertedAI:
@rf-ivtdai @Ruishenl pinging you two for feedback on this PR.

invertedai/utils.py
```diff
@@ -443,17 +443,26 @@ def _interpret_response_line(self, result):
         return data
 
 @validate_call
-def get_default_agent_attributes(agent_count_dict: Dict[str,int]) -> List[AgentAttributes]:
+def get_default_agent_properties(
+    agent_count_dict: Dict[str,int],
+    attribute_type: Optional[str] = "properties"
```
rf-ivtdai, line 448: If this is meant to be a flag to by default use or not use `agent_properties`, change it to boolean `use_agent_properties` with a default value. Also update the doc string below accordingly.

invertedai/large/common.py
```diff
@@ -85,3 +96,14 @@ def is_inside(
         x, y = point.x, point.y
         region_x, region_y = self.center.x, self.center.y
         return region_x - self.size/2 <= x and x <= region_x + self.size/2 and region_y - self.size/2 <= y and y <= region_y + self.size/2
+
+    def convert_attributes_to_properties(self,attributes):
```
rf-ivtdai, line 100: It seems this method could be called in a few other places as well. Consider moving it to utils and call it everywhere you need to convert attributes into properties.

invertedai/large/drive.py
```diff
@@ -86,8 +86,25 @@ def large_drive(
...
-    if not (len(agent_states) == len(agent_attributes) == len(recurrent_states)):
+    if not (len(agent_states) == len(agent_properties) == len(recurrent_states)):
         raise InvalidRequestError(message="Input lists are not of equal size.")
+    if not len(agent_states) > 0:
+        raise InvalidRequestError(message="Valid call must contain at least 1 agent.")
+
+    # Convert any AgentAttributes to AgentProperties for backwards compatibility 
+    agent_properties_new = []
+    for properties in agent_properties:
+        properties_new = properties
```
rf-ivtdai, line 97: call `convert_attributes_to_properties`

invertedai/utils.py
KieranRatcliffeInvertedAI: Done

invertedai/large/common.py
KieranRatcliffeInvertedAI: Done.

KieranRatcliffeInvertedAI:
@rf-ivtdai if the changes I've made from your suggestions look good, is there anything else to change before merging this branch?

Ruishenl (approved):

rf-ivtdai:
> @rf-ivtdai if the changes I've made from your suggestions look good, is there anything else to change before merging this branch?

Looks good :)

rf-ivtdai (approved):

PR #215, opened 2024-08-08
Log replay

KieranRatcliffeInvertedAI:
Added tools to be able to conveniently read and write logs in the python SDK.


KieranRatcliffeInvertedAI:
PLEASE NOTE:
Some changes have been made to the `DriveResponse` and `InitializeResponse` due to an uncommon issue with `recurrent_states` being None. Previously, these 2 different types, despite their close interaction within scripts that implement them, have slightly different types in regards to how they handle the null case (i.e. is it an Optional List or a List of Optionals). 

In this PR, it was chosen to make them consistent towards the Optional List format to line up with how the Drive payload is formatted (i.e. the user doesn't have to check if it's an empty list or a list of None's then change it to be just None so that the call to Drive does not break). 

This spawned a larger discussion between myself and @rf-ivtdai since this has larger implications about our design philosophy since it turns out our Response data structures types have indeed been occupying multiple roles: 1. Capture the response data from server 2. Contain the information needed for the subsequent Drive call. These 2 functions have been the exact same up until this moment when it became useful to produce a `DriveResponse` or `InitializeResponse` variable that contained None for their `recurrent_states` then this dual functionality became evident.

invertedai/logs/common.py
```diff
@@ -0,0 +1,456 @@
...
+            velocity_vec = velocity_vec,
+            plot_frame_number = plot_frame_number
+        )
+
+class LogWriter(LogHandler):
+    """
+    A class for conveniently writing a log to a JSON log format.
+    """
+
+    def __init__(self):
```
Ruishenl, line 131: You don't need to declare it if no explicit modification is needed

invertedai/logs/common.py
```diff
@@ -0,0 +1,456 @@
...
+                scenario_log.rendering_center[0],
+                scenario_log.rendering_center[1]
+            ]
+        }
+
+        with open(log_path, "w") as outfile:
+            json.dump(output_dict, outfile)
+
+    @classmethod
+    def write_json_from_scenario_log(
```
Ruishenl, line 261: This seems to be doing the same thing as `write_scenario_log_to_json` ?  And it doesn't make any sense to pass a cls to an instance method. I am not sure if this will work without passing a concrete object to it

invertedai/logs/common.py
Ruishenl, line 1: This module is called common but you are importing stuff from one level above. This might easily cause circular imports in the future and break the package level.
Also having docstrings for each method would be very helpful

invertedai/logs/common.py
```diff
@@ -0,0 +1,456 @@
...
+    initialize_model_version: Optional[str] = "best" #: Please refer to the documentation of :func:`initialize` for information on the api_model_version parameter.
+    drive_model_version: Optional[str] = "best" #: Please refer to the documentation of :func:`drive` for information on the api_model_version parameter.
+    
+    light_recurrent_states: Optional[LightRecurrentStates] = None #: As of the most recent time step. Please refer to the documentation of :func:`drive` for further information on this parameter.
+    recurrent_states: Optional[List[RecurrentState]] = None #: As of the most recent time step. Please refer to the documentation of :func:`drive` for further information on this parameter.
+
+    waypoints: Optional[Dict[str,List[Point]]] = None #: As of the most recent time step. A list of waypoints keyed to agent ID's not including waypoints already passed. These waypoints are not automatically populated into the agent properties.
+
+
+class LogHandler():
```
Ruishenl, line 48: Maybe LogVisualizer or LogVisualizingInterface is a better name here

KieranRatcliffeInvertedAI: I've gone with LogBase as an alternative name since I felt like that was a bit more clear for the role it's meant to fulfill.

invertedai/logs/common.py
KieranRatcliffeInvertedAI: What name would you suggest? Perhaps logger.py, logging.py, logs.py, log_utils.py, anything else?

invertedai/logs/common.py
KieranRatcliffeInvertedAI: It theoretically doesn't make a difference in this instance but it does initialize the 2 instance parameters. My main reasoning however is that if changes are made in the future or by a user, then they don't need to worry about adding this line to make sure the LogWriter class is initialized properly.

invertedai/logs/common.py
```diff
@@ -0,0 +1,456 @@
...
+    """
+    A class for containing features relevant to both log reading and writing such as visualization.
+    """
+
+    def __init__(self):
+        self.scenario_log = None
+        self.simulation_length = None
+
+    def visualize_up_to_timestep(
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/logs/common.py
KieranRatcliffeInvertedAI: I can add the docstrings for sure.

Ruishenl (approved):
I don't really have much further to add, i will let @rf-ivtdai  to take another look

rf-ivtdai (approved):

invertedai/logs/common.py
KieranRatcliffeInvertedAI: Tested this and it indeed works. I can see it being useful if a user doesn't want to create a LogWriter object just to write the log to disk.

PR #217, opened 2024-09-20
Conditional large initialize

KieranRatcliffeInvertedAI:


KieranRatcliffeInvertedAI:
TODO: 
Add infractions to conditional initialize.

invertedai/large/initialize.py
```diff
@@ -274,29 +240,112 @@ def get_number_of_agents_per_region_by_drivable_area(
...
+    else:
+        region_map = None
+
+    for i, (prop, state) in enumerate(zip(agent_properties[:num_agent_states],agent_states)):
+        region_distances = []
+        for region in regions:
+            region_distances.append(sqrt((state.center.x-region.center.x)**2 + (state.center.y-region.center.y)**2))
+
+        closest_region_index = region_distances.index(min(region_distances))
+        insert_index = len(regions[closest_region_index].agent_states)
```
Ruishenl, line 293: Since you are only appending new agents here you can just add a starting index at the top and incrementing it in each iteration instead of grabbing the length of the list every time.  This is correct but if you eventually decide to alter the insertion order in the loop you need to make sure the index you append to the region_map is up-to-date

invertedai/large/initialize.py
```diff
@@ -274,29 +240,112 @@ def get_number_of_agents_per_region_by_drivable_area(
 
     return filtered_regions
 
+@validate_call
+def _insert_agents_into_nearest_regions(
+    regions: List[Region],
+    agent_properties: List[AgentProperties],
+    agent_states: List[AgentState],
+    is_return_map: Optional[bool] = False,
```
Ruishenl, line 248: I think `return_region_index` is a more suitable name for this parameter here.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: It's a bit more complicated than just appending the agents. Since this is initialize, the lengths of the agent_states and agent_properties lists are not necessarily equal. In this loop, it is inserting agents that have both an AgentState and AgentProperties objects into regions that may have agents without AgentStates. The only way to insert these without previously inserted agents indexes changing is to insert it at the ID that corresponds to the end of the agent_states list.

But nonetheless, just so I understand your suggestion, you're recommending to track the length of the agent_states list in each region and just increment that every time loop based on which region was added to? If so that would make this loop a bit faster at the expensive of some small memory size increase.

invertedai/large/initialize.py
KieranRatcliffeInvertedAI: Fair enough. I tend to start boolean parameters with "is" because it typically makes it clear that it is boolean and what question the True or False is answering. This seems clear enough to me though and sounds better.

invertedai/large/initialize.py
Ruishenl: yes simply a small change that will potentially make things faster.

KieranRatcliffeInvertedAI: Actually this could end up making this function take a longer amount of time if the number of agents to insert is smaller than the number of regions since this loop effectively only starts counting region agent numbers if the conditional initialize feature is used and will unnecessarily count the length of the regions if no conditional agents are provided.

KieranRatcliffeInvertedAI: Okay so I put this region length calculation/incrementing as well as this particular loop under an if statement that checks if the agent_states list parameter is greater than zero to avoid unnecessary calculation if no conditional agents are provided.

Ruishenl (approved):

PR #222, opened 2024-10-11
Scenario in drive

Ruishenl:


invertedai/common.py
```diff
@@ -308,6 +308,21 @@ def fromdict(cls, d):
...
 
+class Scenario(BaseModel):
+    name: str
+
+
+class LeaderFollow(Scenario):
+    name: str =  "leader_follower"
+    leader_id: int
+    follower_id: int
+    gap_in_seconds: float = 1.5
```
adscib, line 319:
```suggestion
```
We don't have a model for this one. I just included it as a formatting example.

PR #225, opened 2024-10-30
Initial commit with logger tool to dump all requests and responses

KieranRatcliffeInvertedAI:
This tool could require some work. For example, writing the log will fail if the location_info and initialize responses/requests are not added properly. This can be fixed but will take more time. For now, if the example is followed, it will write a JSON log that can be read.


KieranRatcliffeInvertedAI:
I've changed the logger so that it collects all requests and responses automatically provided a path environment variable is set to an existing directory.

invertedai_cpp/invertedai/session.h
```diff
@@ -36,6 +39,13 @@ class Session {
   double max_backoff = 0; // No max backoff by default, 0 signifies no limit
   double jitter_factor = 0.5;
 
+  const char *iai_logger_path_char = std::getenv("IAI_LOGGER_PATH");
+  bool is_log_path_null = iai_logger_path_char == NULL;
+  const std::string iai_logger_path = !is_log_path_null ? iai_logger_path_char : "./";
+  const bool is_logging = (!is_log_path_null) && (std::filesystem::is_directory(std::filesystem::path(iai_logger_path)));
```
adscib, line 45: I think we should attempt to create this directory if it doesn't exist, and only report a warning if that fails.

invertedai_cpp/invertedai/logger.cc
```diff
@@ -0,0 +1,131 @@
...
+        // log["drive_requests"] = this->write_iai_data_vector_to_log_<DriveRequest>(this->drive_requests_);
+        // log["drive_responses"] = this->write_iai_data_vector_to_log_<DriveResponse>(this->drive_responses_);
+
+        log["drive_requests"] = this->drive_requests_;
+        log["drive_responses"] = this->drive_responses_;
+
+        log["drive_request_times"] = this->drive_request_times_;
+        log["drive_response_times"] = this->drive_response_times_;
+
+        std::string file_path = "iai_log" + std::to_string(this->get_current_time_in_milliseconds_()) + ".json";
```
adscib, line 125: Maybe we can use `YYYY_MM_DD_HH_MM_SS_xxxx` instead? It will be easier to find the relevant log that way.

adscib (commented):
Looks good! I left a couple of suggestions regarding file handling.

invertedai_cpp/invertedai/logger.cc
```diff
@@ -0,0 +1,131 @@
...
+        std::string full_path = dir_path + file_path;
```
Ruishenl, line 126: If `dir_path` does not end with a slash will this work ?

KieranRatcliffeInvertedAI: Unfortunately it will "work" but everything before the last slash will be added to the name of the file instead of the directory path.

KieranRatcliffeInvertedAI: This issue I think will be resolved by following Adam's suggestion of trying to make the directory if it doesn't exist.

invertedai_cpp/invertedai/logger.cc
KieranRatcliffeInvertedAI: To confirm, this will be just for the log name and keep the UNIX time for the request/response times?

KieranRatcliffeInvertedAI: @Ruishenl what would be most useful in terms of time formatting for the request and response times? UNIX milliseconds or a human readable date down to the millsecond as Adam is suggesting for the file name?

invertedai_cpp/invertedai/session.cc
```diff
@@ -172,9 +183,20 @@ const std::string Session::request(
       is.push(src);
       std::stringstream strstream;
       boost::iostreams::copy(is, strstream);
-      return strstream.str();
+
+      const std::string output = strstream.str();
+      if (this->is_logging){
+        this->logger.append_request(output,mode);
```
Ruishenl, line 189: should this be `append_response` ?
```suggestion
        this->logger.append_response(output,mode);
```

KieranRatcliffeInvertedAI: Yes good catch.

invertedai_cpp/invertedai/logger.cc
Ruishenl: If i understand your code correctly, the current setup will only write to the file when the session is closed ?
Either way i think `YYYY-MM-DDTHH:MM:SS.sssZ` would be the most common one and i would use it for the request/response data, in terms of file name i think using epoch millisecond is okay since it's easier to sort.

invertedai_cpp/invertedai/logger.cc
Ruishenl: Sure as long as you grabbed the correct name with path when it's created.  Or use something like `boost::filesystem::path` it should handle both cases

invertedai_cpp/invertedai/logger.cc
KieranRatcliffeInvertedAI: I think I will go with Adam's suggestion for the file title since our users will need to sift through these logs just as often as we will but just I confirm, you also want the human readable date/time for the timestamps of the request/response data?

Ruishenl: Yeah having human readable UTC  datatime in the data  would be better.

invertedai_cpp/invertedai/session.h
KieranRatcliffeInvertedAI: Fixed

invertedai_cpp/invertedai/logger.cc
KieranRatcliffeInvertedAI: Please check the date_time format I went with. @adscib I started with `YYYY_MM_DD_HH_MM_SS_xxx` but it looked bad so I changed it to `YYYY-MM-DD_HH:MM:SS:xxx`. This can be changed relatively easy so please let me know what you think.

invertedai_cpp/invertedai/logger.cc
KieranRatcliffeInvertedAI: So the functions in the std::filesystem library didn't pad the path with a "/" as expected so I added a line to check if it ended in a "/" and add one if it is not present.

Ruishenl (approved):

Ruishenl:
Since this is a new feature, i bumped up the version string for cpp.

PR #226, opened 2024-11-27
Debug logger

KieranRatcliffeInvertedAI:
Added a default debug logger to the python SDK that will log all requests and responses automatically with a debug log path environment variable is provided.

### DIFF
--- a/invertedai/logs/debug_logger.py
+++ b/invertedai/logs/debug_logger.py
@@ -1,39 +1,39 @@
-from pydantic import BaseModel, validate_arguments
-from typing import List, Optional, Dict, Tuple
-from datetime import datetime
-
+import logging
 import json
+import os
+
+from collections import defaultdict
+from typing import List, Optional, Dict, Tuple
+from datetime import datetime, timezone
+
+logger = logging.getLogger(__name__)
 
 class DebugLogger:
-
-    def __init__(self,debug_log_path):
+    def __init__(
+        self,
+        debug_log_path: str
+    ):
         self.debug_log_path = debug_log_path
+        self._create_directory()
 
-        self.data = {
-            "location_info_requests":[],
-            "location_info_responses":[],
-            "location_info_request_timestamps":[],
-            "location_info_response_timestamps":[],
-            "initialize_requests":[],
-            "initialize_responses":[],
-            "initialize_request_timestamps":[],
-            "initialize_response_timestamps":[],
-            "drive_requests":[],
-            "drive_responses":[],
-            "drive_request_timestamps":[],
-            "drive_response_timestamps":[],
-        }
+        self.data = defaultdict(list)
 
         file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
         self.log_path = self.debug_log_path + file_name
 
+    def _get_current_time_human_readable_UTC(self):
+        return datetime.now(timezone.utc).strftime("%Y-%m-%d_%H:%M:%S:%f")
+
     def _create_directory(self):
         if not os.path.isdir(self.debug_log_path):
+            logger.info(f"Debug log directory does not exist: Created new directory at {self.debug_log_path}")
             os.makedirs(self.debug_log_path)
-            print(f"Debug log directory does not exist: Created new directory at {self.debug_log_path}")
-        print(f"Directory already exists!")
 
-    def append_request(self,model,data_dict):
+    def append_request(
+        self,
+        model: str,
+        data_dict: dict
+    ):
         ts = self._get_current_time_human_readable_UTC()
         data_str = json.dumps(data_dict)
 
@@ -51,7 +51,11 @@ class DebugLogger:
 
         self.write_data_to_log()
 
-    def append_response(self,model,data_dict):
+    def append_response(
+        self,
+        model: str,
+        data_dict: dict
+    ):
         ts = self._get_current_time_human_readable_UTC()
         data_str = json.dumps(data_dict)
 
@@ -68,9 +72,6 @@ class DebugLogger:
             self.data["drive_response_timestamps"].append(ts)
 
         self.write_data_to_log()
-
-    def _get_current_time_human_readable_UTC(self):
-        return datetime.now().strftime("%d-%m-%Y_%H:%M:%S:%f")
 
     def write_data_to_log(self):
         with open(self.log_path, "w") as outfile:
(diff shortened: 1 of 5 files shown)


invertedai/logs/debug_logger.py
```diff
@@ -0,0 +1,77 @@
...
+        file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
+        self.log_path = self.debug_log_path + file_name
+
+    def _create_directory(self):
+        if not os.path.isdir(self.debug_log_path):
+            os.makedirs(self.debug_log_path)
+            print(f"Debug log directory does not exist: Created new directory at {self.debug_log_path}")
+        print(f"Directory already exists!")
+
+    def append_request(self,model,data_dict):
```
Ruishenl, line 36: I am not so sure if "model" is the right name for parameter here. Maybe api_name ? And what do you think of having enum here instead of raw string for these api names.

invertedai/logs/debug_logger.py
```diff
@@ -0,0 +1,77 @@
...
+            self.data["initialize_responses"].append(data_str)
+            self.data["initialize_response_timestamps"].append(ts)
+
+        elif model == "drive":
+            self.data["drive_responses"].append(data_str)
+            self.data["drive_response_timestamps"].append(ts)
+
+        self.write_data_to_log()
+
+    def _get_current_time_human_readable_UTC(self):
```
Ruishenl, line 72: I don't think this will give you UTC by default ? Did you set it anywhere else ?

invertedai/logs/debug_logger.py
```diff
@@ -0,0 +1,77 @@
...
+            "drive_request_timestamps":[],
+            "drive_response_timestamps":[],
+        }
+
+        file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
+        self.log_path = self.debug_log_path + file_name
+
+    def _create_directory(self):
+        if not os.path.isdir(self.debug_log_path):
+            os.makedirs(self.debug_log_path)
```
Ruishenl, line 32: i think we should use logger.info and logger.warning here instead of print

invertedai/utils.py
```diff
@@ -243,13 +245,20 @@ def request(
         self, model: str, params: Optional[dict] = None, data: Optional[dict] = None
     ):
         method, relative_path = iai.model_resources[model]
+        
+        if self._debug_logger is not None:
+            self._debug_logger.append_request(model,json.dumps(data))
```
Ruishenl, line 250: So you are passing a json string here into the append_request() but this function itself is accepting a dictionary as well and later dumps it again ?

invertedai/utils.py
```diff
@@ -243,13 +245,20 @@ def request(
...
+
         response = self._request(
             method=method,
             relative_path=relative_path,
             params=params,
             json_body=data,
         )
 
+        if self._debug_logger is not None:
+            self._debug_logger.append_response(model,json.dumps(response))
```
Ruishenl, line 260: Same for this one

invertedai/logs/debug_logger.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import List, Optional, Dict, Tuple
+from datetime import datetime
+
+import json
+
+class DebugLogger:
+
+    def __init__(self,debug_log_path):
+        self.debug_log_path = debug_log_path
+
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: It is possible for some endpoints to not be called in edge cases but since this is a debug log and will be mainly used for visual observation, I think this will be fine to not necessarily explicitly have the log contain these keys.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: "model" is the name used in the "utils.py" Session class which is where this log is used so I named this parameter to align with that naming structure. I can change that in Session as well if you think that'll be worthwhile.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: You're right, I'll make that change.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: Will do

invertedai/utils.py
KieranRatcliffeInvertedAI: Unfortunately this is done to align with the C++ SDK where it writes requests/responses as strings which are appended into the C++ equivalent of lists and then the whole thing needs to be written to a JSON file.

KieranRatcliffeInvertedAI: Also to clarify, what is being passed to these functions is a Dict object which is then dumped into a json string.

KieranRatcliffeInvertedAI: I'll add data types to the function parameters to clarify this.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: I do like the idea of using an enum but I'm planning on making some changes to the log format that will impact both python and C++ SDK however this branch was meant to only replicate the C++ debug log into python.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: Done

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: Done.

invertedai/logs/debug_logger.py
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Done.

invertedai/logs/debug_logger.py
```diff
@@ -0,0 +1,78 @@
...
+    def _create_directory(self):
+        if not os.path.isdir(self.debug_log_path):
+            logger.info(f"Debug log directory does not exist: Created new directory at {self.debug_log_path}")
+            os.makedirs(self.debug_log_path)
+
+    def append_request(
+        self,
+        model: str,
+        data_dict: dict
```
Ruishenl, line 35: Not really important but the correct typing here and below should be `Optional[Dict]=None`

Ruishenl (approved):

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: In what situations would data be `None`? I think that change would be fine but I'm asking in case I'm wrong and there is more that should be done to handle that `None` case.

PR #231, opened 2024-12-16
Added a fix to ensure location_info requests are captured accurately …

KieranRatcliffeInvertedAI:
…in the log.

### DIFF
--- a/invertedai/utils.py
+++ b/invertedai/utils.py
@@ -262,9 +262,8 @@ class Session:
         method, relative_path = iai.model_resources[model]
         
         if self._debug_logger is not None:
-            if data is not None:
-                request_data = data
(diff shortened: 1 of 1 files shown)


Ruishenl (approved):
I actually find your previous commit more future proof. If later we end up adding another api that takes query parameters we will need to update this line as well.

KieranRatcliffeInvertedAI:
> I actually find your previous commit more future proof. If later we end up adding another api that takes query parameters we will need to update this line as well.

Perhaps but it is unclear what the difference is between "data" and "params" and why is it different in location_info? If I knew that more for sure then I could change the logic to be indeed future proof while being careful of committing to something based on an accident.

PR #235, opened 2024-12-21
Update Cosimulation Functionality

KieranRatcliffeInvertedAI:
Changed the python examples folder to be compatible with updates in the SDK.


invertedai/cosimulation.py
```diff
@@ -1,114 +1,75 @@
...
         self,
         location: str,
-        ego_agent_mask: Optional[List[bool]] = None,
-        monitor_infractions: bool = False,
-        get_birdview: bool = False,
-        random_seed: Optional[int] = None,
-        traffic_lights: bool = False,
-        # sufficient arguments to initialize must also be included
-        **kwargs,
+        ego_agent_properties: Optional[List[AgentProperties]] = None,
```
Ruishenl, line 45: Are`ego_agent_properties` and `ego_agent_agent_states` bascially the conditional agents information passed down to INITIALIZE ?
Shouldn't `conditional_` be a better prefix than `ego_` here ?

Also i notice the new interface doesn't allow us to use a separate mask to decide which agents to control anymore, instead it is expecting the user to control all the passed ego agents. Would there be a case where they pre-define some agents but only want to control a subset of them ?

Ruishenl (commented):
There seems to be some major interface changes in cosimulation.py, i took a look at it and seems to be okay, but ideally we should add some simple tests to make sure we don't accidenlty break it in the future. But that can be a separate PR.

invertedai/cosimulation.py
```diff
@@ -1,114 +1,82 @@
...
         self,
         location: str,
-        ego_agent_mask: Optional[List[bool]] = None,
-        monitor_infractions: bool = False,
-        get_birdview: bool = False,
-        random_seed: Optional[int] = None,
-        traffic_lights: bool = False,
-        # sufficient arguments to initialize must also be included
-        **kwargs,
+        conditional_agent_properties: Optional[List[AgentProperties]] = None,
```
rf-ivtdai, line 50: We are obviously switching to a completely different backend process i.e. large_initialize and large_drive, which means completely different ways to create the same objects, and dropping support for certain properties. There are probably very little usage on this, but do we have a process for "backward breaking" changes like these on the client side?

PR #236, opened 2025-01-08
Debug Log Visualizer Feature

KieranRatcliffeInvertedAI:


invertedai/logs/debug_logger.py
```diff
@@ -75,4 +79,173 @@ def append_response(
...
+
+        return scene_plotter, response_data
+
+    @classmethod
+    def visualize_log(
+        cls,
+        log_data: Dict,
+        gif_name: str = "./debug_log_visualization.gif",
+        fov: int = 100,
+        map_center: Tuple[float] = None
```
Ruishenl, line 149: I think the more accurate annotation here should be `Tuple[float, float`

invertedai/logs/debug_logger.py
```diff
@@ -75,4 +79,173 @@ def append_response(
...
+        )
+        plt.close(fig)
+
+    @classmethod
+    def reproduce_log(
+        cls,
+        log_data: Dict,
+        gif_name: str = "./debug_log_reproduction.gif",
+        fov: int = 100,
+        map_center: Tuple[float] = None,
```
Ruishenl, line 180: same here

Ruishenl (approved):

PR #240, opened 2025-02-14
Diagnostic tool

KieranRatcliffeInvertedAI:
Added a new tool that can be expanded upon for user-side quick diagnostic checks to highlight common issues when experiencing difficulties.

### DIFF
--- a/invertedai/logs/debug_logger.py
+++ b/invertedai/logs/debug_logger.py
@@ -14,25 +14,38 @@ import matplotlib.pyplot as plt
 logger = logging.getLogger(__name__)
 
 class DebugLogger:
+    def __new__(cls, *args, **kwargs):
+        if not cls.check_instance_exists(cls):
+            cls.instance = super(DebugLogger, cls).__new__(cls)
+        return cls.instance
+
+    def check_instance_exists(cls):
+        return hasattr(cls, 'instance')
+
     def __init__(
         self,
-        debug_log_path: str
+        debug_dir_path: Optional[str] = None
     ):
-        self.debug_log_path = debug_log_path
-        self._create_directory()
+        if debug_dir_path is not None:
+            self.debug_dir_path = debug_dir_path
+            self._create_directory()
 
-        self.data = defaultdict(list)
+            self.data = defaultdict(list)
 
-        file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
-        self.log_path = self.debug_log_path + file_name
+            self.init_time = datetime.timestamp(datetime.now())
+            file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
+            self.debug_log_path = os.path.join(self.debug_dir_path,file_name)
+
+    def reinitialize_logger(self):
+        self.__init__(debug_dir_path = self.debug_dir_path)
 
     def _get_current_time_human_readable_UTC(self):
         return datetime.now(timezone.utc).strftime("%Y-%m-%d_%H:%M:%S:%f")
 
     def _create_directory(self):
-        if not os.path.isdir(self.debug_log_path):
-            logger.info(f"Debug log directory does not exist: Created new directory at {self.debug_log_path}")
-            os.makedirs(self.debug_log_path)
+        if not os.path.isdir(self.debug_dir_path):
+            logger.info(f"Debug log directory does not exist: Created new directory at {self.debug_dir_path}")
+            os.makedirs(self.debug_dir_path)
 
     def append_request(
         self,
@@ -53,6 +66,14 @@ class DebugLogger:
         elif model == "drive":
             self.data["drive_requests"].append(data_str)
             self.data["drive_request_timestamps"].append(ts)
+
+        elif model == "large_initialize":
+            self.data["large_initialize_requests"].append(data_str)
+            self.data["large_initialize_request_timestamps"].append(ts)
+
+        elif model == "large_drive":
+            self.data["large_drive_requests"].append(data_str)
+            self.data["large_drive_request_timestamps"].append(ts)
 
         self.write_data_to_log()
 
@@ -76,10 +97,18 @@ class DebugLogger:
             self.data["drive_responses"].append(data_str)
             self.data["drive_response_timestamps"].append(ts)
 
+        elif model == "large_initialize":
+            self.data["large_initialize_responses"].append(data_str)
+            self.data["large_initialize_response_timestamps"].append(ts)
+
+        elif model == "large_drive":
+            self.data["large_drive_responses"].append(data_str)
+            self.data["large_drive_response_timestamps"].append(ts)
+
         self.write_data_to_log()
 
     def write_data_to_log(self):
-        with open(self.log_path, "w") as outfile:
+        with open(self.debug_log_path, "w") as outfile:
             json.dump(self.data, outfile)
 
     def _get_scene_plotter(
--- a/invertedai/logs/debug_logger.py
+++ b/invertedai/logs/debug_logger.py
@@ -23,7 +23,6 @@ class DebugLogger:
 
         self.data = defaultdict(list)
 
-        self.init_time = datetime.timestamp(datetime.now())
         file_name = "iai_log_" + self._get_current_time_human_readable_UTC() + "_UTC.json"
         self.debug_log_path = os.path.join(self.debug_dir_path,file_name)
 
--- a/invertedai/logs/debug_logger.py
+++ b/invertedai/logs/debug_logger.py
@@ -14,6 +14,15 @@ import matplotlib.pyplot as plt
 logger = logging.getLogger(__name__)
 
 class DebugLogger:
+    """
+    A tool for capturing debug logs which contain all serialized data from every request and response.
+
+    Parameters
+    ----------
+    debug_log_path:
+        The full path to the debug log directory where debug logs are written.
+    """
+
     def __init__(
         self,
         debug_dir_path: Optional[str] = None
@@ -190,6 +199,22 @@ class DebugLogger:
         fov: int = 100,
         map_center: Tuple[float,float] = None
     ):
+        """
+        Visualize a debug log using the SDK visualizer tools. This assumes the debug log was captured in chronological order
+        by simulated time step.
+
+        Parameters
+        ----------
+        log_data:
+            A loaded JSON dictionary of a debug log.
+        gif_name:
+            The path and name of the resulting GIF file.
+        fov:
+            The field of view in metres for the birdview visualization.
+        map_center:
+            The coordinates within the map on which to centre the visualization which is especially useful for large maps.
+        """
+
         scene_plotter, _ = cls._get_scene_plotter(
             cls,
             log_data=log_data,
@@ -222,6 +247,25 @@ class DebugLogger:
         map_center: Tuple[float,float] = None,
         use_log_seed: bool = True
     ):
+        """
+        Given the initial state captured in the debug log and using all relevant information, attempt to reproduce the simulation
+        from the debug log and visualize it for comparison with the original. This assumes the debug log was captured in chronological order
+        by simulated time step.
+
+        Parameters
+        ----------
+        log_data:
+            A loaded JSON dictionary of a debug log.
+        gif_name:
+            The path and name of the resulting GIF file.
+        fov:
+            The field of view in metres for the birdview visualization.
+        map_center:
+            The coordinates within the map on which to centre the visualization which is especially useful for large maps.
+        use_log_seed:
+            A flag for whether to use the random seed in the debug log or input a value of None to DRIVE.
+        """
+
         scene_plotter, response_data = cls._get_scene_plotter(
             cls,
             log_data=log_data,
@@ -273,6 +317,19 @@ class DebugLogger:
         is_reproduce_log: bool = False,
         **kwargs
     ):
+        """
+        A convenient entry point for quickly visualizing and/or reproducing a given debug log.
+
+        Parameters
+        ----------
+        log_data:
+            A loaded JSON dictionary of a debug log.
+        is_visualize_log:
+            A flag to control whether the log is visualized. Please refer to the appropriate function for the relevant keyword arguments.
+        is_reproduce_log:
+            A flag to control whether the log is reproduced. Please refer to the appropriate function for the relevant keyword arguments.
+        """
+
         with open(debug_log_path) as json_file:
             log_data = json.load(json_file)
 
--- a/invertedai/logs/debug_logger.py
+++ b/invertedai/logs/debug_logger.py
@@ -14,28 +14,18 @@ import matplotlib.pyplot as plt
 logger = logging.getLogger(__name__)
 
 class DebugLogger:
-    def __new__(cls, *args, **kwargs):
-        if not cls.check_instance_exists():
-            cls.instance = super(DebugLogger, cls).__new__(cls)
-        return cls.instance
-
-    @classmethod
-    def check_instance_exists(cls):
-        return hasattr(cls, 'instance')
-
     def __init__(
         self,
         debug_dir_path: Optional[str] = None
     ):
-        if debug_dir_path is not None:
-            self.debug_dir_path = debug_dir_path
-            self._create_directory()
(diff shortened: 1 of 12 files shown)


invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,105 @@
...
+        for req_dict, res_dict in zip(req_groupings["recurrent_states"][1:],res_groupings["recurrent_states"][:-1]):
+            is_equal_recurrent_states.append(self._check_states_equal(req_dict,res_dict))
+
+        if not all(is_equal_agent_states):
+            res_states = list(filter(lambda i: is_equal_agent_states[i], range(len(is_equal_agent_states))))
+            diagnostic_messages.append(f"Potential agent state index error detected at log timestep(s): {res_states}")
+
+        if not all(is_equal_recurrent_states):
+            res_states = list(filter(lambda i: is_equal_recurrent_states[i], range(len(is_equal_recurrent_states))))
+            diagnostic_messages.append(f"Potential recurrent state index error detected at log timestep(s): {res_states}")
```
adscib, line 60: In a typical use case, the client is inserting modified actions for the ego agent, which would always get flagged here. I would suggest two modifications to make it easier to sort out:
1. Print the indices of agents affected by the problem.
2. Add an argument to the diagnostic tool that lets you specify the indices of ego agents, for which the checks will be omitted. Make it default to `[0]`.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,105 @@
...
+
+        return req_agent_state_dict, res_agent_state_dict
+
+    def _check_individual_states_equal(self,sa_0,sa_1):
+        for param_0, param_1 in zip(sa_0,sa_1):
+            if not param_0 == param_1: return False
+
+        return True
+
+    def _check_states_equal(self,states_0,states_1):
```
adscib: I think you should account for agents appearing and disappearing in between calls. It's probably reasonable to assume that agents can be dropped arbitrarily, but when they're introduced they're always appended to the end of the list.

adscib (changes requested):
Nice. The structure makes sense, but I think we need to tweak the warning logic to make it useful.

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: These are both good ideas and I'll add them.

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: This could be good to look out for that it might be intentional that an agent is dropped between time steps. However, user intent is going to be impossible to assume during these diagnostics and an agent missing in the next time step _could_ be a mistake or might result in other issues like removing the wrong corresponding recurrent state.

I think one thing to keep in mind that a diagnostic tool like this has a lower standard of certainty which is why I've worded the messages as "Potential issue...". So even if the "issue" detected isn't an issue in reality, pointing out where things _might_ have gone wrong is the idea and then the user can judge for themselves the validity. 

That being said, a high false-positive rate will indeed lower trust in the tool so the level of specificity will be something that needs to be balanced.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+            res_states = [f"({j}:{is_equal_recurrent_states["agents_equal"][j]})" for j, val in enumerate(is_equal_recurrent_states["is_equal"]) if not val]
+            diagnostic_messages.append(f"Potential recurrent state index error detected for (Time step:[Agent Indexes]): {res_states}")
+
+        return diagnostic_messages
+
+    def _get_all_drive_agents(
+        self,
+        log_data: Dict
+    ):
+        STATE_SIGDIG = 2
```
Ruishenl, line 74: Naming wise since you are using it with round(), i believe they are number of decimal places not significant digit

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+                state_dict["recurrent_states"].append([[round(x,RECURR_SIGDIG) for x in st] for st in data["recurrent_states"]])
+
+        return req_agent_state_dict, res_agent_state_dict
+
+    def _check_individual_states_equal(
+        self,
+        sa_0: float,
+        sa_1: float
+    ):
+        for param_0, param_1 in zip(sa_0,sa_1):
```
Ruishenl, line 103: Why zip here ?

invertedai/logs/debug_logger.py
```diff
@@ -16,23 +16,27 @@
...
-        debug_log_path: str
+        debug_dir_path: Optional[str] = None
     ):
-        self.debug_log_path = debug_log_path
+        self.debug_dir_path = debug_dir_path
         self._create_directory()
 
         self.data = defaultdict(list)
 
+        self.init_time = datetime.timestamp(datetime.now())
```
Ruishenl, line 26: i could be wrong but is this being used anywhere ?

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+            states_equal, is_state_equal = self._check_states_equal(req_dict,res_dict)
+            is_equal_agent_states["agents_equal"].append(states_equal)
+            is_equal_agent_states["is_equal"].append(is_state_equal)
+        for req_dict, res_dict in zip(req_groupings["recurrent_states"][1:],res_groupings["recurrent_states"][:-1]):
+            states_equal, is_state_equal = self._check_states_equal(req_dict,res_dict)
+            is_equal_recurrent_states["agents_equal"].append(states_equal)
+            is_equal_recurrent_states["is_equal"].append(is_state_equal)
+
+        if not all(is_equal_agent_states["is_equal"]):
+            res_states = [f"({j}:{is_equal_agent_states["agents_equal"][j]})" for j, val in enumerate(is_equal_agent_states["is_equal"]) if not val]
```
Ruishenl, line 61: CI is broken because you use double quotes again in the f string

Ruishenl (changes requested):
If you already have some test cases i would recommend turn them into unit tests and put them in a new file under tests/

invertedai/logs/diagnostics.py
rf-ivtdai: Debug mode is broken for the same reason..

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+            diagnostic_messages.append(f"Potential agent state index error detected for (Time step:[Agent Indexes]): {res_states}")
+
+        if not all(is_equal_recurrent_states["is_equal"]):
+            res_states = [f"({j}:{is_equal_recurrent_states["agents_equal"][j]})" for j, val in enumerate(is_equal_recurrent_states["is_equal"]) if not val]
```
rf-ivtdai, line 65: Same as above

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+
+        print(f"Finished diagnostic analysis.")
+
+    def _check_drive_response_equivalence(self):
+        diagnostic_messages = []
+
+        is_equal_agent_states = {"is_equal":[],"agents_equal":[]}
+        is_any_agent_states_issue = False
+        is_equal_recurrent_states = {"is_equal":[],"agents_equal":[]}
+        is_any_recurrent_states_issue
```
rf-ivtdai, line 47: Unfinished here?

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+        self,
+        log_data: Dict
+    ):
+        STATE_SIGDIG = 2
+        RECURR_SIGDIG = 6
+
+        req_agent_state_dict = {"agent_states":[],"recurrent_states":[]}
+        res_agent_state_dict = {"agent_states":[],"recurrent_states":[]}
+
+        is_large_simulation = len(log_data["large_drive_responses"]) > 0
```
rf-ivtdai, line 80: When only regular iai.drive is called, you will get a Key Error on "large_drive_responses".

KieranRatcliffeInvertedAI: Fixed

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: Fixed

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: Fixed

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: Fixed

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: You're right, this function was reduced until it was not necessary. Fixed.

invertedai/logs/debug_logger.py
KieranRatcliffeInvertedAI: Thank you for catching that. This was also a layover from a previous stage of development.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+            drive_req_data = log_data["drive_requests"]
+            drive_res_data = log_data["drive_responses"]
+
+        for req_data, res_data in zip(drive_req_data,drive_res_data):
+            res_data = json.loads(res_data)
+            req_data = json.loads(req_data)
+
+            for (state_dict, data) in zip([req_agent_state_dict,res_agent_state_dict],[req_data,res_data]):
+                state_dict["agent_states"].append([[round(x,STATE_SIGDIG) for x in st] for st in data["agent_states"]])
+                state_dict["recurrent_states"].append([[round(x,RECURR_SIGDIG) for x in st] for st in data["recurrent_states"]])
```
rf-ivtdai, line 94: There is a valid workflow that the initial drive call does not pass `recurrent_state`. NoneType needs to be handled properly here.

KieranRatcliffeInvertedAI: Yes you are right, good call.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+        req_agent_state_dict = {"agent_states":[],"recurrent_states":[]}
+        res_agent_state_dict = {"agent_states":[],"recurrent_states":[]}
+
+        is_large_simulation = len(log_data["large_drive_responses"]) > 0
+        if is_large_simulation:
+            drive_req_data = log_data["large_drive_requests"]
+            drive_res_data = log_data["large_drive_responses"]
+        else:
+            drive_req_data = log_data["drive_requests"]
+            drive_res_data = log_data["drive_responses"]
```
rf-ivtdai, line 86: When a drive request triggers a properly handled exception, such as `InvalidRequestError`, `drive_responses` will also be absent from the debug log. This is probably true for `large_drive_responses` as well. We will need to handle both 2xx successful responses, and 4xx or even 5xx ones.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+        Parameters
+        ----------
+        debug_log_path:
+            The full path to the debug log file to be loaded and analyzed.
+        """
+
+        self.debug_log_path = debug_log_path
+        self.log_data = None
+        with open(debug_log_path) as json_file:
+            self.log_data = json.load(json_file)
```
rf-ivtdai, line 26: A minor thing is when user sets IAI_LOGGER_PATH and misses the ending '/', our code demands to be babysat :p, loading with os.path.join that combines the logger base path and log name might be better.

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: So far this tool accepts debug logs and those logs only capture the request and response data, not errors. We could make that change since theoretically an error could be missed if the user is not paying attention which would then be highlighted in the diagnostic run.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,140 @@
...
+            is_equal_agent_states["agents_equal"].append(states_equal)
+            is_equal_agent_states["is_equal"].append(is_state_equal)
+        for req_dict, res_dict in zip(req_groupings["recurrent_states"][1:],res_groupings["recurrent_states"][:-1]):
+            states_equal, is_state_equal = self._check_states_equal(req_dict,res_dict)
+            is_equal_recurrent_states["agents_equal"].append(states_equal)
+            is_equal_recurrent_states["is_equal"].append(is_state_equal)
+
+        if not all(is_equal_agent_states["is_equal"]):
+            res_states = [f"({j}:{is_equal_agent_states["agents_equal"][j]})" for j, val in enumerate(is_equal_agent_states["is_equal"]) if not val]
+            diagnostic_messages.append(f"Potential agent state index error detected for (Time step:[Agent Indexes]): {res_states}")
```
rf-ivtdai, line 62: I would propose to make the printed message easier to read here. For a happy-path, 60s simulation, the user sees a screen of True and False, per timestep per agent. It will be might easier to see in the format of: "{type of potential issue} : {iterate on timestep}, {which agent has this issue}.

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: Fixed

invertedai/logs/diagnostics.py
KieranRatcliffeInvertedAI: I made the default for the ego agents to an empty list since I feel it is more sensible to default towards analyzing more rather than less and it will be very obvious to the user if they forgot to include the ego agents list.

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,215 @@
...
+        self,
+        timestep: int,
+        issue_type: str,
+        agent_list: Optional[str] = None
+    ):
+        if agent_list is None:
+            output_str = f"Potential issue that might cause degradation in performance at timestep {timestep}: {issue_type}"
+        else:
+            output_str = f"Potential issue that might cause degradation in performance at timestep {timestep}: {issue_type} applicable to agents {agent_list}"
+        return output_str
```
rf-ivtdai, line 50: Much better now! I shall be that annoying guy again - We might as well return something like: For each time step, here are the issue codes and all their applicable agents. This will still print as many lines as there are timesteps, so it might be worth it to allow users to "hide" messages from certain timesteps by specifying an index of timesteps to show/hide? Textual explanation of the issue codes are included once at the bottom, could be exactly the definition dictionary.

KieranRatcliffeInvertedAI: I see the logic here. It is not guaranteed that if there is an issue present, that it will occur every time step though the chances are high that it will be systematic and produce messages every step so fair enough. One issue that I can spot is that the agent index relevant to the issues might change every time step which fits less nicely into this print scheme. My initial thought is to follow your recommendation here but print out the relevant agent IDs for each time step along with the issue code.

rf-ivtdai: The default option would be to include all timesteps, which definitely "prints the most complete" picture :) Compared to potentially hiding timesteps, I believe the more sensible simplification here is showing the issue code instead of full text for each line, and just include the legend at the bottom.

KieranRatcliffeInvertedAI: So then don't include the relevant agent IDs which will likely change every time step in this simplified version?

rf-ivtdai: Sometimes when the agent IDs change between timesteps, that by itself is an indication for the issue at hand?

invertedai/logs/diagnostics.py
```diff
@@ -0,0 +1,354 @@
...
+                state_dict["recurrent_states"].append(recurr_state)
+
+        for req_json in drive_req_data:
+            req_data = json.loads(req_json)
+            req_agent_details.append(self._get_agent_details(req_data))
+        
+        if "large_initialize_responses" in log_data:
+            init_res_data = log_data["large_initialize_responses"]
+        else:
+            init_res_data = log_data["initialize_responses"]
```
rf-ivtdai, line 302: It is also valid when user does not call initialize and only call drive.

KieranRatcliffeInvertedAI: Fixed.

PR #255, opened 2025-03-28
Added capability for agents to be removed or added in various SDK tools

KieranRatcliffeInvertedAI:
Also added multiple waypoint functionality to the logger tool.

### DIFF
--- a/invertedai/logs/logger.py
+++ b/invertedai/logs/logger.py
@@ -179,8 +179,36 @@ class LogWriter(LogBase):
         used to export a given scenario log instead of the log contained within the object.
         """
 
+        individual_suggestions_dict = {}
         if scenario_log is None:
             scenario_log = self._scenario_log
+            for i, prop in enumerate(scenario_log.agent_properties):
+                wp = prop.waypoint
+                if wp is not None:
+                    individual_suggestions_dict[str(i)] = {
+                        "suggestion_strength": 0.8,
+                        "states": {
+                            "0": {
+                                "center": {
+                                    "x": wp.x,
+                                    "y": wp.y
+                                }
+                            }
+                        }
+                    }
+        else:
+            if scenario_log.waypoints is not None:
+                for agent_id, wps in scenario_log.waypoints.items():
+                        individual_suggestions_dict[agent_id] = {
+                            "suggestion_strength": 0.8,
+                            "states": [{
+                                "center": {
+                                    "x": wp.x,
+                                    "y": wp.y
+                                }
+                            } for wp in wps]
+                        }
+
         num_cars, num_pedestrians = 0, 0
         for prop in scenario_log.agent_properties:
             if prop.agent_type == "car":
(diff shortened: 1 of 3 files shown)


invertedai/logs/logger.py
```diff
@@ -49,6 +49,7 @@ class ScenarioLog(BaseModel):
 
     waypoints: Optional[Dict[str,List[Point]]] = None #: As of the most recent time step. A list of waypoints keyed to agent ID's not including waypoints already passed. These waypoints are not automatically populated into the agent properties.
 
+    present_mask: Optional[List[List[int]]] = None #: List of indexes corresponding to agent_properties for which agents are present at each time step. If None, all agents are present at every time step.
```
Ruishenl, line 52: I usually would expect a mask to be boolean instead of int. By reading the comments looks like this is more like `present_agents` pr `present_agent_indices` ? I don't think this is used as mask in your code.

invertedai/utils.py
```diff
@@ -60,6 +60,8 @@
     500: "The server encountered an unexpected issue. We're working to resolve this. Please try again later.",
 }
 
+Color = Tuple[float,float,float]
+ColorList = List[Optional[Color]]
```
Ruishenl, line 64: Does it mean it can be something like [None, None] ?

Ruishenl (changes requested):

invertedai/logs/logger.py
KieranRatcliffeInvertedAI: You're right, fixed the language to be more clear.

invertedai/utils.py
KieranRatcliffeInvertedAI: Correct, in this cause it uses the default colour.

invertedai/logs/logger.py
```diff
@@ -381,6 +426,20 @@ def drive(
 
         self.simulation_length += 1
 
+    def get_present_agents(self):
```
Ruishenl, line 429:
```suggestion
    def get_present_agent_indexes(self): 
```

invertedai/logs/logger.py
```diff
@@ -381,6 +426,20 @@ def drive(
 
         self.simulation_length += 1
 
+    def get_present_agents(self): 
+        """
+        Returns which agents are currently present within the simulation.
```
Ruishenl, line 431:
```suggestion
        Returns the indexes of the agents that are currently present within the simulation.
```

invertedai/logs/logger.py
```diff
@@ -179,8 +185,36 @@ def export_to_file(
...
         """
 
+        individual_suggestions_dict = {}
         if scenario_log is None:
             scenario_log = self._scenario_log
+            for i, prop in enumerate(scenario_log.agent_properties):
+                wp = prop.waypoint
+                if wp is not None:
+                    individual_suggestions_dict[str(i)] = {
+                        "suggestion_strength": 0.8,
```
Ruishenl, line 195: Maybe we should add a comment to explain this hard-coded value ?

invertedai/logs/logger.py
KieranRatcliffeInvertedAI: Done

KieranRatcliffeInvertedAI:
@Ruishenl the last commit moved the present_indexes functionality in the logger tool to the ScenarioLog data type itself.

Ruishenl (approved):

PR #258, opened 2025-04-29
Upgrade numpy and carla to latest version

Ruishenl:
We can do a dev release first to test the new package to see if it install correctly @KieranRatcliffeInvertedAI


KieranRatcliffeInvertedAI:
For visibility, we will need to figure out the required `numpy` version for the Carla python API before I can approve this PR. Last I checked they need before 2.x so we'll need to double check and speak with their dev team if that's still the case.

Ruishenl:
> For visibility, we will need to figure out the required `numpy` version for the Carla python API before I can approve this PR. Last I checked they need before 2.x so we'll need to double check and speak with their dev team if that's still the case.

The fact that i am able to auto resolve both numpy 2.x and carla 0.9.15 is a promising sign

pyproject.toml
```diff
@@ -13,14 +13,13 @@ packages = [
 python = ">=3.10,<3.13"
 requests = "^2.28.1"
 pydantic = "^2.5.3"
-matplotlib = "^3.6.2"
 tqdm = "^4.65.0"
-numpy = "^1.24.4"
-
+numpy = "^2.2.5"
```
KieranRatcliffeInvertedAI, line 17: Carla has yet to increment their version of numpy therefore I cannot justify this PR at this time.

KieranRatcliffeInvertedAI (changes requested):
I'm going to select "Request changes" at this time though the effort is on Carla's side on whether they're going to increment their version of Numpy.

PR #259, opened 2025-06-09
Added example of AV2.0 pipeline for adding agents to a scene builder …

KieranRatcliffeInvertedAI:
…scenario.

Added an example showing a basic example of adding background agents to a log following an AV2.0 pipeline.

### DIFF
--- a/invertedai/logs/logger.py
+++ b/invertedai/logs/logger.py
@@ -500,7 +500,7 @@ class LogReader(LogBase):
         all_agent_properties_unsorted = {}
         present_indexes_unsorted = []
         agent_id_list = {}
-        curr_agent_index = 0
+        agent_id_sequence_num = 0
         for i in range(LOG_DATA["scenario_length"]):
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,7 +3,7 @@ dev/
 dev.py
 img/
 gifs/
-examples/output
```
Ruishenl, line 6: why do we want to keep the output sub folder ?

invertedai/logs/logger.py
```diff
@@ -499,26 +499,29 @@ def __init__(
         all_agent_states_unsorted = []
         all_agent_properties_unsorted = {}
         present_indexes_unsorted = []
+        agent_id_list = {}
+        curr_agent_index = 0
```
Ruishenl, line 503: Is this only for counting the sequence number for each agent_id ? Maybe `agent_id_sequence_num` is a better name ?

PR #262, opened 2025-09-20
Large_Initialize

atao-inverted:


invertedai_cpp/invertedai/data_utils.h
```diff
@@ -344,6 +374,57 @@ struct AgentProperties {
...
+//                 // if you still support AgentAttributes fallback
+//                 AgentProperties props;
+//                 props.agent_type = agent_type_to_string(agent_type);  //only type
+//                 agent_attributes_list.push_back(props);
+//             }
+//         }
+//     }
+
+//     return agent_attributes_list;
+// }
```
KieranRatcliffeInvertedAI: Production quality code should not contain commented out blocks. Please delete.

invertedai_cpp/invertedai/data_utils.h
```diff
@@ -8,11 +8,41 @@
...
+};
+
+// Helper to convert to string
+inline std::string to_string(AgentType type) {
+    switch (type) {
+        case invertedai::AgentType::car: return "car";
+        case invertedai::AgentType::pedestrian: return "pedestrian";
+        default: return "unknown";
+    }
+}
```
KieranRatcliffeInvertedAI: This function seems like a good candidate for the "inline" keyword but I don't think this function is used frequently enough to warrant the downsides.

invertedai_cpp/invertedai/data_utils.h
```diff
@@ -8,11 +8,41 @@
...
+
+// Helper to parse from string (if needed)
+inline invertedai::AgentType agent_type_from_string(const std::string& s) {
+    if (s == "car") return invertedai::AgentType::car;
+    if (s == "pedestrian") return invertedai::AgentType::pedestrian;
+    throw std::invalid_argument("Unknown AgentType: " + s);
+}
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: Furthermore this function does not appear to be used outside of commented out functions so please delete this and all other functions that are not used.

invertedai_cpp/large/common.h
```diff
@@ -0,0 +1,158 @@
...
+//         const Point2d& center,
+//         double size = 100.0,
+//         const std::vector<invertedai::AgentState>& states = {},
+//         const std::vector<invertedai::AgentProperties>& props = {},
+//         const std::vector<invertedai::RecurrentState>& recurs = {}
+//     ) {
+//         if (states.size() != props.size()) {
+//             throw std::invalid_argument("states and props must have same length");
+//         }
+//         if (!recurs.empty() && recurs.size() != states.size()) {
```
KieranRatcliffeInvertedAI: This file doesn't appear to have any actual code in here. Please delete this file.

invertedai_cpp/large/large_helpers.h
```diff
@@ -0,0 +1,167 @@
+    #pragma once
+
+    #include <vector>
+    #include <map>
+    #include <optional>
+    #include <random>
+    #include <stdexcept>
+    #include "common.h"          // Region, Point, AgentType, etc.
```
KieranRatcliffeInvertedAI: Delete reference to "common.h" if the file is to be deleted.

invertedai_cpp/sstitched_with_agents.png
KieranRatcliffeInvertedAI: Please make sure no images are included in production quality code base unless it is meant to go into the README. Add *.png files to the .gitignore file.

invertedai_cpp/large/quadtree.h
KieranRatcliffeInvertedAI: This file is not meant to be in this PR, please remove it.

invertedai_cpp/large/larger_helpers.cc
KieranRatcliffeInvertedAI: Why do both this file and large_init_helpers.cc/.h exist? Also largeR_helpers is an odd name and should be changed if there's a good reason for this file to exist when the other helper file exists.

invertedai_cpp/invertedai/data_utils.h
```diff
@@ -408,14 +489,127 @@ struct StaticMapActor {
...
+      double size = 100.0,
+      const std::vector< AgentState>& states = {},
+      const std::vector< AgentProperties>& props = {},
+      const std::vector<std::vector<double>>& recurs = {}
+  ) {
+      if (states.size() != props.size()) {
+          throw std::invalid_argument("states and props must have same length");
+      }
+      if (!recurs.empty() && recurs.size() != states.size()) {
+          throw std::invalid_argument("recurs must be empty or same length as states");
```
KieranRatcliffeInvertedAI: When writing error messages, think of how an external developer will read it. The messages should be informative as well as professionally worded. Please change this to something like "The recurrent states vector must be empty or the same length as the number of given agents."

invertedai_cpp/invertedai/data_utils.h
```diff
@@ -408,14 +489,127 @@ struct StaticMapActor {
...
+}
+void insert_all_agent_details(
+  const AgentState& state,
+  const AgentProperties& props,
+  const std::vector<double>& rec
+) {
+  agent_states.push_back(state);
+  agent_properties.push_back(props);
+  recurrent_states.push_back(rec);
+}
```
KieranRatcliffeInvertedAI: Please maintain a consistent indentation and spacing scheme to make code more readable. The spacing of your function parameters is better but these member functions need to be indented to show that they are indeed members of the above struct.

invertedai_cpp/large/main.cpp
```diff
@@ -0,0 +1,373 @@
...
+using namespace invertedai;
+
+// Clamp helper
+inline int clampi(int v, int lo, int hi) { return std::max(lo, std::min(v, hi)); }
+
+// ensure all regions are 100m x 100m 
+auto validate_regions_100x100 (const std::vector<invertedai::Region>& regs,
+    double expected = 100.0,
+    double tol = 1e-6,
+    double px_per_meter = -1.0) {
```
KieranRatcliffeInvertedAI: Using this as an example but please make the spacing on the function definitions consistent and compliant with the following standard:

image

Please update ALL function definitions in all files to follow this standard. The only exception is if there is 1 or fewer arguments to the function.

invertedai_cpp/large/main.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (ok) {
+        std::cout << "[OK] All " << regs.size()
+        << " regions are 100m x 100m (FOV=" << expected << ")\n";
+    }
+    return ok;
+    };
+
+// Simple agent generator for testing
+    inline std::pair<std::vector<AgentState>, std::vector<AgentProperties>>
+    generate_agents_for_region(const Region& region,
```
KieranRatcliffeInvertedAI: Can you explain to me the reason for this function as it seems quite similar to a regular initialize call?

invertedai_cpp/sstitched_with_agents.png
atao-inverted: Okay I will make sure for future commits this will not happen again

invertedai_cpp/large/main.cpp
atao-inverted: Very good point. I will swap it out for a regular initialize call. The regular initialize call is definitely the better choice here.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+            int rand_idx = dist(rng);
+            regions[rand_idx].agent_properties.push_back(agent_properties[i]);
+    
+            if (return_region_index) {
+                region_map.emplace_back(rand_idx, static_cast<int>(regions[rand_idx].agent_properties.size() - 1));
+            }
+        }
+    
+        return {regions, region_map};
+    }
```
KieranRatcliffeInvertedAI: Please fix the indentation here.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+        regions[closest_idx].agent_properties.insert(
+            regions[closest_idx].agent_properties.begin() + insert_index, prop);
+
+        if (return_region_index) {
+            region_map.emplace_back(closest_idx, insert_index);
+        }
+    }
+    
+    
+        // 2. Place remaining properties (agents without states)
```
KieranRatcliffeInvertedAI: What numbered list is this comment referring to?

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+    if (i == exclude_index) continue;
+
+    const Region& r = regions[i];
+    if (std::sqrt(squared_distance(nearby_region.center, r.center)) >
+        (REGION_MAX_SIZE + AGENT_SCOPE_FOV_BUFFER)) {
+        continue;
+    }
+    const auto n = std::min(r.agent_states.size(), r.agent_properties.size());
+    out_states.insert(out_states.end(), r.agent_states.begin(), r.agent_states.begin() + n);
+    out_props.insert(out_props.end(), r.agent_properties.begin(), r.agent_properties.begin() + n);
```
KieranRatcliffeInvertedAI: It is a lot cleaner to explicitly return objects compared to modifying them within a function. Unless there is a good reason, please only pass objects by value instead of reference to ensure they're immutable within the function.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+    }
```
KieranRatcliffeInvertedAI: Please fix the indentation of this for loop.

KieranRatcliffeInvertedAI (changes requested):

invertedai_cpp/large/large_initialize.cc
atao-inverted: I also want to rename my variables for clairty. I will do that right now.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,596 @@
...
+            throw std::invalid_argument("Not enough agent properties for the given agent states.");
+        }
+
+        // find nearest region                                   
+        double min_dist = std::numeric_limits<double>::max();
+        int closest_idx = -1;
+        Point2d agent_center{state.x, state.y};
+        for(size_t r = 0; r< num_regions; r++) {
+            const auto& region = regions[r];
+            double dist = squared_distance(agent_center, region.center);
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: There appears to be a miscommunication. What I meant to ask was why this function that is meant to calculate L2 distance was not taking the square root which would be how the L2 distance is calculated? 

Disregard what I said about using abs(), that was confusing and by no means should the L1 distance be used. 

But my original question still stands, why is the computed distance not square rooted, especially when the square root is taken on line 152 anyways?

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+            region_map.emplace_back(closest_idx, insert_index);
+        }
+    }
+    
+    
+        // 2. Place remaining properties (agents without states)
+        std::mt19937 rng(random_seed.value_or(std::random_device{}()));
+        std::uniform_int_distribution<int> dist(0, static_cast<int>(num_regions - 1));
+    
+        for (size_t i = num_agent_states; i < agent_properties.size(); i++) {
```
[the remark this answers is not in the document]
atao-inverted: oh wow that is very true, that would definitely make it safer for the future.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+        std::vector<AgentProperties> fov_props;
+        fov_states.reserve(existing_states.size());
+        fov_props.reserve(existing_props.size());
+        for (size_t j = 0; j < existing_states.size(); ++j) {
+            Point2d agent_center{existing_states[j].x, existing_states[j].y};
+            if (inside_fov(region_center, region.size + AGENT_SCOPE_FOV_BUFFER, agent_center)) {
+                fov_states.push_back(existing_states[j]);
+                fov_props.push_back(existing_props[j]);
+            }
+        }
```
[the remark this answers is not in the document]
atao-inverted: Oops sorry, this is dead code I forgot to delete. I will delete it now.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+                        if (!props.max_speed)        props.max_speed = (type == "car" ? 15.0 : 1.5);
+                        if (!props.waypoint)         props.waypoint = Point2d{0,0};
+                    } else {
+                        // must NOT include static geometry → strip down to type + optional waypoint
+                        props.length.reset();
+                        props.width.reset();
+                        props.rear_axis_offset.reset();
+                        props.max_speed.reset();
+                        // keep agent_type and waypoint only
+                    }
```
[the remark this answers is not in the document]
atao-inverted: Yes very true. This is also old debugging code I should have deleted as I was trying to pinpoint past errors.  I will delete it now.

invertedai_cpp/large/large_initialize.cc
```diff
@@ -0,0 +1,588 @@
...
+                if (!return_exact_agents) {
+                    Point2d agent_center{state.x, state.y};
+                    if (!inside_fov(region_center, region.size, agent_center)) {
+                        continue;
+                    }
+                }
+                if (r_state.size() > 1000) {
+                    std::cerr << "[ERROR] Region " << i
+                                << " recurrent[" << j << "] has absurd size="
+                                << r_state.size() << " (skipping)" << std::endl;
```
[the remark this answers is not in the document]
atao-inverted: This was definitely apart of old debugging code that is now irrelevant. I will go through all my std::cerr messages right now just to check if they are still relevant.

invertedai_cpp/large/BUILD
KieranRatcliffeInvertedAI: Looking at the file structure here and given how many more files will be included for large_drive as well, I think it might be useful to create a subdirectory under large called large_initialize that contains all relevant files. Let me know if you think this would cause any practical issues for sourcing.

invertedai_cpp/large/large_init_helpers.cc
```diff
@@ -0,0 +1,221 @@
...
+            filtered_regions.push_back(region);
+        }
+    }
+
+    if (total_drivable_area_ratio <= 0.0) {
+        throw InvertedAIError("No drivable area found in any region.");
+    }
+
+    return filtered_regions;
+    }
```
KieranRatcliffeInvertedAI: Fix the indentation of these functions here please.

invertedai_cpp/large/main.cpp
KieranRatcliffeInvertedAI: If this is meant to be the example showing how to implement large_initialize AND the visualizer tool, please separate these into separate files or at least functions within this executable. Imagine yourself as a prospective developer wanting to understand how to use this function, they might have a hard time parsing between these two functionalities or worse, get discouraged because it looks complicated when in reality there are 2 significant chunks of functionality here.

PR #266, opened 2025-11-05
Waypoint visualizer

atao-inverted:


invertedai/utils.py
```diff
@@ -1095,7 +1105,7 @@ def animate_scene(
         fig.set_size_inches(self._resolution[0] / self._dpi, self._resolution[1] / self._dpi, True)
 
         def animate(i):
-            self._update_frame_to(i)
+            self._update_frame_to(i, waypoints=waypoints_per_frame if waypoints_per_frame is not None else None)
```
KieranRatcliffeInvertedAI: You can just pass waypoints_per_frame because it is already either `None` or the correct data type.

invertedai/utils.py
```diff
@@ -1220,15 +1234,28 @@ def _get_color(
 
         return c
 
-    def _update_frame_to(self, frame_idx):
+    def _update_frame_to(self, frame_idx, waypoints:Optional[List[List[Point]]] = None):
         for rect in self.actor_boxes.values():
             rect.set_visible(False)
+        if hasattr(self, "waypoint_markers"):
```
KieranRatcliffeInvertedAI: This check is unnecessary since the class is by default given this attribute.

invertedai/logs/logger.py
```diff
@@ -106,7 +106,8 @@ def visualize_range(
         velocity_vec: bool = False,
         plot_frame_number: bool = True,
         left_hand_coordinates: bool = False,
-        agent_ids: Optional[List[int]] = None
+        agent_ids: Optional[List[int]] = None,
+        waypoints_per_frame: Optional[List[List[Point]]] = None
```
KieranRatcliffeInvertedAI: It may be a good idea to allow the user to insert the waypoints per time step as that is how the state data is passed into this visualizer. Please keep data entry consistent.

KieranRatcliffeInvertedAI (approved):

invertedai/logs/logger.py
```diff
@@ -106,7 +106,8 @@ def visualize_range(
         velocity_vec: bool = False,
         plot_frame_number: bool = True,
         left_hand_coordinates: bool = False,
-        agent_ids: Optional[List[int]] = None
+        agent_ids: Optional[List[int]] = None,
+        waypoints_per_frame: Optional[List[Dict[int, Optional[List[float]]]]] = None
```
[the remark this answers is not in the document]
KieranRatcliffeInvertedAI: I guess my question here is why is the value in the key-value pair of this Dict optional? This would be the case if we needed to list every agent in `waypoints_per_frame` therefore need to cover the case where an agent has no waypoint. But why not just not include that agent in the Dict if it doesn't have a waypoint during that frame?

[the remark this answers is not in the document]
atao-inverted: I see, 

The only reason I made the Point optional is for consistency with the way it is represented with the `CarlaWaypoints` class, this way I could simply do a `waypoints_per_frame.append(given_waypoints_from_background_traffic) ` at each `drive() `step and append the carla waypoints given from `background_traffic.py`

But I see the value in representing more simply. I suppose for the visualization we could make it `Dict[int, Point]` and then when adding the data in, filtering out the '`None`' waypoint values. So the data entry logic would be a little more complicated but this change should be fine

KieranRatcliffeInvertedAI:
@steven9909 please provide a review as soon as you can so that we can merge this branch and you can merge this into your working branch.

steven9909 (approved):
Looks good!

PR #269, opened 2025-12-03
Add lane change option when generating lane ids

steven9909:


invertedai/helpers/waypoints.py
```diff
@@ -1,24 +1,184 @@
...
     """
-    Generates a list of waypoints from a sequence of lane ids.
+    Generates a list of waypoints from a sequence of lane ids. The start state should be within the first lane.
 
     Args:
         start_state (AgentState): The starting state of the agent.
         lanelet_map (lanelet2.core.LaneletMapLayers): Projected lanelet map.
         lane_ids (List[int]): Sequence of lane ids to follow.
         waypoint_spacing (float): Spacing between the waypoints in meters. Defaults to 15.
+        waypoint (Optional[Point], optional): Desired final waypoint. Defaults to None.
```
bzwartsenberg: Can you name this "destination_waypoint"? It gets very confusing otherwise.

invertedai/helpers/waypoints.py
```diff
@@ -75,13 +269,13 @@ def get_lanelet(id):
...
 
     return waypoints
-    
 
 def generate_lane_ids_from_lanelet_map(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
     target_distance: float = 600.0, 
-    waypoint: Optional[Point] = None
+    waypoint: Optional[Point] = None,
```
bzwartsenberg: Same here. And anywhere else I am not aware of.

invertedai/helpers/waypoints.py
```diff
@@ -135,12 +328,11 @@ def generate_lane_ids_from_lanelet_map(
         lane_length = lanelet2.geometry.length2d(current_lanelet)
         path.append(current_lanelet.id)
         total_lane_distance += lane_length
-        reachable_lanelets = routing_graph.following(current_lanelet, withLaneChanges=False)
+        reachable_lanelets = routing_graph.following(current_lanelet, withLaneChanges=lane_change)
```
bzwartsenberg: Is this the same as including `routing_graph.left` and `routing_graph.right` in the `following` call?

We need to make sure there's a cost associated with a lane change, so there are no random lane changes if they're not necessary I think.

bzwartsenberg: Also, does this return the lanes that are adjacent to each other?

invertedai/helpers/waypoints.py
steven9909: Resolved!

invertedai/helpers/waypoints.py
steven9909: Resolved!

invertedai/helpers/waypoints.py
steven9909: It returns the lanes adjacent to each other if the lane subtypes are "dashed"

invertedai/helpers/waypoints.py
```diff
@@ -124,74 +319,74 @@ def get_lanelet(id):
...
 
     Returns:
-        List[int]: Sequence of lane ids to follow.
+        List[int]: Sequence of lane ids to follow. Empty if no routes are possible.
     """
-    if target_distance is None:
-        target_distance = 600.0
-    traffic_rules = lanelet2.traffic_rules.create(lanelet2.traffic_rules.Locations.Germany,
-                                                    lanelet2.traffic_rules.Participants.Vehicle)
+    random.seed(seed)
```
lioutasb: You should use [random generators](https://numpy.org/devdocs/reference/random/generator.html) instead of setting the global seed.

invertedai/helpers/waypoints.py
```diff
@@ -124,74 +319,74 @@ def get_lanelet(id):
...
 
 def generate_lane_ids_from_lanelet_map(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
-    target_distance: Optional[float] = None ,
-    waypoint: Optional[Point] = None
+    min_distance: Optional[float] = None, 
+    destination_waypoint: Optional[Point] = None,
+    lane_change: bool = False,
+    seed: int = 0,
```
lioutasb: The seed should be None by default, letting the RNG pick one.

invertedai/helpers/waypoints.py
```diff
@@ -124,74 +319,74 @@ def get_lanelet(id):
...
 
     return waypoints
-    
 
 def generate_lane_ids_from_lanelet_map(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
-    target_distance: Optional[float] = None ,
-    waypoint: Optional[Point] = None
+    min_distance: Optional[float] = None,
```
lioutasb, line 173: If `min_distance` is optional, then we should not assume a fixed value inside the method.

invertedai/helpers/waypoints.py
```diff
@@ -1,16 +1,170 @@
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Callable
 import lanelet2
 import random
 import numpy as np
+import logging
 
+from scipy.interpolate import interp1d
```
KieranRatcliffeInvertedAI: This package is meant to be as lightweight as possible. `scipy` is a hell of a dependency to include and I would suggest finding a lighter-weight method to do 1D interpolation.

invertedai/helpers/waypoints.py
```diff
@@ -1,16 +1,170 @@
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Callable
 import lanelet2
```
KieranRatcliffeInvertedAI, line 2: Also I missed this on the previous PR but please include this dependency in this package by specifying it in `pyproject.toml`.

invertedai/helpers/waypoints.py
```diff
@@ -49,25 +203,32 @@ def get_default_waypoints(
...
 
 def generate_waypoints_from_lane_ids(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
     lane_ids: List[int], 
-    waypoint_spacing: float = 15.0
+    waypoint_spacing: float = 15.0,
+    destination_waypoint: Optional[Point] = None,
+    transition_distance: int = 3,
+    lane_change_fn: Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray], np.ndarray] = hermite_spline,
```
KieranRatcliffeInvertedAI: There are 2 reasons I can see to make these values parameters instead of hard coded.
1. We want to expose this to users.
2. We call this function in multiple places where the parameters are different.

I don't think either of these apply and if we expose these as parameters, we may be forced into supporting this forever. Please remove these as parameters unless there is a compelling argument to keep them this way,

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: Agreed.

invertedai/helpers/waypoints.py
```diff
@@ -124,74 +319,74 @@ def get_lanelet(id):
...
-    
 
 def generate_lane_ids_from_lanelet_map(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
-    target_distance: Optional[float] = None ,
-    waypoint: Optional[Point] = None
+    min_distance: Optional[float] = None, 
+    destination_waypoint: Optional[Point] = None,
+    lane_change: bool = False,
```
KieranRatcliffeInvertedAI: Same comment as above. For any functions that are being exposed, we need to be careful what parameters we choose due to being committed to supporting them into the future. `destination_waypoint` is fine but `min_distance` and `lane_change` should not be exposed because it is very easy to change how we do lane changes in the future. 

Hard code these values for now. Also why is lane_change set to False by default? Shouldn't we want to the vehicle to change lanes if it needs to?

KieranRatcliffeInvertedAI (commented):
Any functions with comment blocks will be automatically documented when we push this to the documentation branch. I do not think ALL of these functions are worthy of being entry points. Please remove all comment blocks for the smaller functions that we don't want being exposed in the documentation which allows us to modify how they work in the future. Also, this is not obligatory but I happen to prefer it as a style preference, add a `_` to the beginning of function names that are not meant to be public.

Once this is done, please see my other comments before approval.

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: This applies to `transition_distance` and `lane_change_fn` to be clear. Maybe even `waypoint_spacing` too.

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: Agreed, this will affect much more than this line intends to.

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: Actually `waypoint_spacing` should stay as a parameter because I need it.

invertedai/helpers/waypoints.py
steven9909: This package is already in pyproject.ml

invertedai/helpers/waypoints.py
steven9909: Using numpy interp instead

invertedai/helpers/waypoints.py
steven9909: I think the transition_distance might have to stay since the user might want to make it shorter/longer depending on how the maps are structured. If the transition distance is hard coded to be too short, there might be cases where it never reaches the desired waypoint because it couldn't transition in time. I agree with the lane_change_fn though.

PR #271, opened 2025-12-11
Waypoint Manager tool to seamlessly control waypoints

KieranRatcliffeInvertedAI:
Functionality:

1. Instantiate agent properties with waypoint lists.
2. Update these lists checking for whether waypoints have been achieved.
3. Reroute agents (re-instantiate the waypoints lists) if the full path has been achieved.
4. Be as stateless as possible (to handle changing agents).
5. Be as lightweight and unintrusive as possible.

### DIFF
--- a/invertedai/helpers/waypoints.py
+++ b/invertedai/helpers/waypoints.py
@@ -81,7 +81,8 @@ def generate_lane_ids_from_lanelet_map(
     start_state: AgentState, 
     lanelet_map: lanelet2.core.LaneletMapLayers, 
     target_distance: float = 600.0, 
-    waypoint: Optional[Point] = None
+    waypoint: Optional[Point] = None,
+    lane_change: bool = False
 ) -> List[int]:
     """
     Generates a sequence of lane ids. If given a waypoint, it will generate the shortest possible route between
@@ -93,6 +94,7 @@ def generate_lane_ids_from_lanelet_map(
         lanelet_map (lanelet2.core.LaneletMapLayers): Projected lanelet map.
         target_distance (float): Target distance in meters to generate. Ignored if waypoint is specified. Defaults to 600.
         waypoint (Optional[Point], optional): Desired final waypoint. Defaults to None.
+        lane_change (bool): Whether lane changes are supported. Defaults to False.
 
     Returns:
         List[int]: Sequence of lane ids to follow.
@@ -119,7 +121,7 @@ def generate_lane_ids_from_lanelet_map(
         possible_routes = []
         for _, ending_lanelet in ending_lanelets:
             for starting_lanelet in filtered_lanelets:
-                possible_route = routing_graph.getRoute(starting_lanelet, ending_lanelet, withLaneChanges=False)
+                possible_route = routing_graph.getRoute(starting_lanelet, ending_lanelet, withLaneChanges=lane_change)
                 if possible_route:
                     possible_routes.append(possible_route)
         if not possible_routes:
@@ -135,7 +137,7 @@ def generate_lane_ids_from_lanelet_map(
         lane_length = lanelet2.geometry.length2d(current_lanelet)
         path.append(current_lanelet.id)
         total_lane_distance += lane_length
-        reachable_lanelets = routing_graph.following(current_lanelet, withLaneChanges=False)
+        reachable_lanelets = routing_graph.following(current_lanelet, withLaneChanges=lane_change)
         if reachable_lanelets:
             current_lanelet = random.choice(reachable_lanelets)
         else:
--- a/invertedai/helpers/waypoints.py
+++ b/invertedai/helpers/waypoints.py
@@ -65,7 +65,7 @@ class WaypointManager:
             self.logger.setLevel(self.cfg.log_level)
             self.logger.propagate = False
 
-            self.debug_data = List[List[WaypointManagerLogState]] if self.cfg.log_level is not None else None
+        self.debug_data = List[List[WaypointManagerLogState]] if self.cfg.log_level is not None else None
 
     def update(
         self,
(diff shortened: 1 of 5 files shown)


invertedai/helpers/waypoints.py
```diff
@@ -1,10 +1,100 @@
...
+
+
+        return _agent_properties
+
+    def check_waypoint_achieved(
+        self,
+        agent_state: AgentState,
+        waypoint: Point
+    ) -> bool:
+        return sqrt(waypoint.x - agent_state.center.x) ** 2 + (waypoint.y - agent_state.center.y) ** 2 < self.waypoint_threshold
```
steven9909: I think you missed a bracket here :D

KieranRatcliffeInvertedAI: Indeed yes

invertedai/helpers/waypoints.py
```diff
@@ -1,10 +1,100 @@
...
+                if props.waypoints is None or len(props.waypoints) == 0: #Check both if is None or empty
+                    props.waypoints = generate_waypoints_from_lane_ids(
+                        start_state=state,
+                        lanelet_map=self.lanelet_map, 
+                        waypoint_spacing=self.cfg.waypoint_spacing,
+                        lane_ids=generate_lane_ids_from_lanelet_map(
+                            start_state=state, 
+                            lanelet_map=self.lanelet_map,
+                            # waypoint=destination,
+                            # target_distance=dist
```
KieranRatcliffeInvertedAI: @steven9909 does target distance really need to be a parameter? I'm struggling to think of a use case where this is the best solution vs a default value or another solution like rerouting after achieving the path?

Also we need to have a solution for when a randomly selected waypoint does not have a solution (e.g. on highway maps with no overpasses).

steven9909: I think yeah target_distance is meaningless once we have the rerouting implemented. I think together with rerouting, it can just be removed.

KieranRatcliffeInvertedAI:
@bzwartsenberg @steven9909 @lioutasb  The work on this branch is complete barring any comments. This branch furthermore has merged in the branch in this [PR](https://github.com/inverted-ai/invertedai/pull/269) so merging this branch into develop will effectively merge both. Please leave your reviews and/or approvals so that this PR can be merged as necessary.

invertedai/helpers/waypoints.py
```diff
@@ -1,16 +1,269 @@
...
+        _agent_properties = [AgentProperties.deserialize(props.serialize()) for props in agent_properties]
+        for i, mask in enumerate(agents_mask):
+            props = _agent_properties[i]
+
+            if mask or target_paths[i] is not None:
+                state = agent_states[i]
+                
+                if props.waypoints is None:
+                    # The current agents waypoints need to be initialized
+                    print(f"Generating new waypoints for agent_{i}")
```
KieranRatcliffeInvertedAI: We definitely should not include print statements by default in a function that is meant to be used frequently. Messages to the terminal should be reserved for warnings, errors, etc. and even then should be formatted differently. Print statements in example or demo files are okay but not core tools. Please remove these.

steven9909: Accidentally added these :cry:

invertedai/helpers/waypoints.py
```diff
@@ -1,89 +1,325 @@
...
     Returns:
         List[Point]: List of waypoints for the agent to follow.
     """
-    assert len(lane_ids) >= 1, "Expected the lane_ids to be populated"
+    if len(lane_ids) < 1:
+        msg = f"Cannot find path for agent with state: {start_state}"
+        if logger is not None: 
+            logger.error(msg)
+        else:
+            raise Exception(msg)
```
KieranRatcliffeInvertedAI: @steven9909 is it necessary for this to be an error state or could we just return no waypoints? Or is this a usability question?

invertedai/helpers/waypoints.py
```diff
@@ -107,9 +343,49 @@ def get_lanelet(id):
...
+                    if logger is not None: logger.log("The starting position is behind the first lane centerline point even after adjustment. This may lead to unexpected behavior.")
+            lane_centerline_points.insert(0, Point(x=x, y=y))
+        else:
+            if prev_lanelet:
+                if not current_lanelet in routing_graph.following(prev_lanelet, withLaneChanges=True):
+                    msg = f"Current lanelet not in routing graph."
+                    if logger is not None: 
+                        logger.error(msg)
+                    else:
+                        raise Exception(msg)
```
KieranRatcliffeInvertedAI: @steven9909 I changed this from an assert that didn't have a message to this block that uses the logger object. What is an appropriate error message and does this necessarily need to be an error state?

invertedai/helpers/waypoints.py
steven9909: I think this could be resolved with what we talked about 2 days ago, we could have the user give fallback options if it failed to generate any lane ids, so one option could be, no waypoints, another could be, it just continues on whatever lane it is currently in, etc..

invertedai/helpers/waypoints.py
steven9909: This should be an error state, if the lane ids aren't following each other, there is no way we can route correctly.

KieranRatcliffeInvertedAI: Sounds good, what is a reasonable error message here then because the placeholder I provided is not very helpful.

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: A fail soft feature has been added now

invertedai/helpers/waypoints.py
steven9909: Hmm probably something like "Provided lanes were not a valid sequence of lanes"? Something along that line

PR #273, opened 2025-12-18
C++ LogWriter (ScenarioLogWriter, ScenePlotter) Functionality and Examples

atao-inverted:


invertedai_cpp/invertedai/visualize.cc
```diff
@@ -42,8 +42,8 @@ void ScenePlotter::draw_traffic_lights(
         if (!actor) continue;
         double length_m = std::max(1.0, actor->length.value_or(1.0));
         double width_m = std::max(1.0, actor->width.value_or(1.0));
-        double pixel_length = length_m * 3.5;
-        double pixel_width = width_m * 3.5;
+        double pixel_length = length_m * 875/this->li_res_.rendering_fov();
```
atao-inverted, line 45: these numbers 875 and 625 may seem random but i did the math and they roughly end up as a 3.5 and 2.5 scaling factor that is scaled depending on the map fov

invertedai_cpp/large/large_drive/large_drive.h
```diff
@@ -36,7 +36,7 @@ struct LargeDriveConfig {
     /**
      * @brief Logger for optional API request/response tracking.
      */
-    LogWriter logger;
+    DebugLogger logger;
```
atao-inverted, line 39: old Logwriter -> DebugLogger

invertedai_cpp/invertedai/visualize.h
```diff
@@ -54,6 +53,7 @@ class ScenePlotter {
         bool flip_x_;
         std::vector<StaticMapActor> static_actors_;
         std::map<std::string, cv::Point> traffic_light_positions_;
+        LocationInfoResponse li_res_;
```
atao-inverted, line 55: since i added a rendering_fov field into the LocationInfoResponse we can now keep track of fov using the LocationInfoResponse itself

invertedai_cpp/invertedai/visualize.cc
```diff
@@ -96,11 +96,12 @@ void ScenePlotter::draw_agent(
 
 ScenePlotter::ScenePlotter(
     const LocationInfoResponse& li_res, 
-    int fov,
```
atao-inverted, line 99: since i added a rendering_fov field into the LocationInfoResponse we can now keep track of fov using the LocationInfoResponse itself

invertedai_cpp/invertedai/logger.h
```diff
@@ -97,36 +94,52 @@ namespace invertedai {
             std::vector<AgentProperties> get_agent_properties();
             std::vector<std::vector<AgentState>> get_agent_states_over_time();
             std::optional<std::vector<std::map<std::string, std::string>>> get_traffic_lights_states_over_time();
-
     };
-    class LogWriter {
```
atao-inverted, line 102: moved to its own file

invertedai_cpp/invertedai/debug_logger.h
```diff
@@ -0,0 +1,35 @@
+#include <iostream>
+#include <vector>
+#include <string>
+
+namespace invertedai {
+class DebugLogger {
```
atao-inverted, line 6: this used to be the old OG LogWriter

invertedai_cpp/invertedai/location_info_response.cc
```diff
@@ -25,6 +26,11 @@ LocationInfoResponse::LocationInfoResponse(const std::string &body_str) {
     this->body_json_["map_origin"][0],
     this->body_json_["map_origin"][1]
   };
+  this->rendering_center_ = {
```
atao-inverted, line 29: assuming map_center and map_fov is always available in the API response body...

invertedai_cpp/invertedai/location_info_response.h
```diff
@@ -19,6 +19,8 @@ class LocationInfoResponse {
   std::vector<unsigned char> birdview_image_;
   std::string osm_map_;
   Point2d map_origin_;
+  Point2d rendering_center_;
```
atao-inverted, line 22: In python these are represented as Point and float respectively.

I was debating between using `double` and `float` for the `rendering_fov`, but decided on `float` due to the fact we do not need the precision that comes from a `double`.

invertedai_cpp/examples/scenario_log_example.cc
```diff
@@ -89,15 +149,29 @@ int main(int argc, char** argv) {
     std::vector<AgentProperties> agent_properties = log_reader.current_agent_properties();
     std::optional<std::map<std::string,std::string>> tl_states = log_reader.current_traffic_lights();
     std::optional<std::vector<LightRecurrentState>> light_rnn = log_reader.current_light_recurrent_state();
-    std::optional<std::vector<RecurrentState>> rnn_opt = log_reader.current_recurrent_states();
+    std::optional<std::vector<std::vector<double>>> rnn_opt = log_reader.current_recurrent_states();
```
KieranRatcliffeInvertedAI, line 169: This appears to be consistent with how we define recurrent states in other places in our cpp SDK so this change is approved.

invertedai_cpp/invertedai/debug_logger.cc
KieranRatcliffeInvertedAI, line 1: Can you confirm whether this was copy/pasted directly or if anything else was changed? I only see a wall of green so I don't know if anything else was added at a glance.

KieranRatcliffeInvertedAI: Same question for the header file.

invertedai_cpp/invertedai/location_info_response.cc
KieranRatcliffeInvertedAI: That is a reasonable assumption for map_center but looking at the SDK, I cannot guarantee map_fov is always available.

invertedai_cpp/invertedai/location_info_response.h
KieranRatcliffeInvertedAI: Correct, float is sufficient accuracy.

invertedai_cpp/invertedai/session.h
```diff
@@ -13,9 +13,9 @@
 #include <boost/beast/http.hpp>
 #include <boost/beast/ssl.hpp>
 #include <boost/filesystem.hpp>
-
+#include <filesystem>
```
KieranRatcliffeInvertedAI, line 16: Why has this dependency been added?

invertedai_cpp/invertedai/visualize.cc
KieranRatcliffeInvertedAI: Can you explain why the significance of 3.5 or 2.5 as a scaling factor (independent of the FOV)?

invertedai_cpp/invertedai/location_info_response.cc
atao-inverted: are there any default values we can safely fall back to?

invertedai_cpp/invertedai/visualize.cc
atao-inverted: this was through visual testing

invertedai_cpp/invertedai/debug_logger.cc
atao-inverted: I copy pasted everything

invertedai_cpp/invertedai/session.h
atao-inverted: Let me look into this, because on the develop branch everything works fine

invertedai_cpp/invertedai/logger.cc
```diff
@@ -345,360 +340,431 @@ namespace invertedai {
...
+
+        json light_recurrent_array = json::array();
+        if (log.light_recurrent_states.has_value()) {
+            for (const auto& lrs : log.light_recurrent_states.value()) {
+                // Each entry should be a two-element array: [state, time_remaining]
+                light_recurrent_array.push_back({lrs.state, lrs.time_remaining});
             }
-            
+        } else {
+            std::cout << "[EXPORT] Found no light recurrent states" << std::endl;
```
KieranRatcliffeInvertedAI: Is there a reason [EXPORT] is added to this output message?

KieranRatcliffeInvertedAI: Also is this message necessary? Unless the user can choose a log level (e.g. debug), status messages like this should not be included so as to minimize interference with the user's command line.

invertedai_cpp/invertedai/logger.cc
```diff
@@ -345,360 +340,431 @@ namespace invertedai {
...
+            };
+            output_dict["rendering_centers"] = {
+                log.rendering_center->first,
+                log.rendering_center->second
+            };
         }
+    
+        std::ofstream outfile(log_path);
+        if (!outfile.is_open()) {
+            throw std::runtime_error("Failed to open file: " + log_path);
```
KieranRatcliffeInvertedAI, line 625: An error message like this is correct and necessary as opposed to status messages.

invertedai_cpp/invertedai/logger.cc
atao-inverted: I understand your point. This makes sense to me, thanks.

invertedai_cpp/examples/scenario_log_example.cc
```diff
@@ -31,6 +32,8 @@ using namespace invertedai;
 
 */
 const int TIMESTEP_TO_BRANCH_FROM = 10;
+const int SIMULATION_LENGTH = 50;
+const std::string LOCATION = "carla:Town10HD"; // location used to write the json log
```
KieranRatcliffeInvertedAI: Please make this a command line parameter so that we don't have to rebuild every time we want to test a different map.

KieranRatcliffeInvertedAI (changes requested):

invertedai_cpp/examples/scenario_log_example.cc
atao-inverted: This makes sense. Thank you.

invertedai_cpp/invertedai/visualize.cc
KieranRatcliffeInvertedAI: Right but what is the significance of the numbers? What are they scaling exactly?

atao-inverted: the length and width of the traffic light status bar rendered on the map

atao-inverted: This bar must be scaled relative to the fov, otherwise if their values were static, we would end up with too large or too small traffic light rectangles once we change maps or change fov

KieranRatcliffeInvertedAI (approved):
This PR is good to merge.

PR #274, opened 2025-12-19
Fix waypoints setting

steven9909:


invertedai/helpers/waypoints.py
```diff
@@ -501,33 +501,28 @@ def generate_lane_ids_from_lanelet_map(
...
+        total_distance = 0
+        while ending_lanelets and total_distance <= (min_distance if min_distance is not None else 0):
+            ending_lanelet = rng.choice(ending_lanelets)
+            possible_route = routing_graph.getRoute(starting_lanelet, ending_lanelet, withLaneChanges=True)
+            if not possible_route:
+                continue
+            starting_lanelet = ending_lanelet
+            ending_lanelets = sorted(list(routing_graph.reachableSet(starting_lanelet, maxRoutingCost=1000.0, allowLaneChanges=True)), key=lambda lanelet: lanelet.id)
+            total_distance += possible_route.length2d()
+            route.extend([lanelet.id for lanelet in list(possible_route.shortestPath())[1:]])
```
KieranRatcliffeInvertedAI, line 526: I thought the main point of this PR was to delete this second block of code and just sample a destination point in the map and reuse the code block above. Is there a reason that cannot be done? It was my understanding that points could be sampled within the lanelets on the map.

steven9909: It is using the getRoute function, but since this function only returns a list of lane ids, I dont think it makes sense to sample a point, but rather sample a lane to go to

PR #275, opened 2026-01-06
Add sampling weights based on number of simultaneous lane changes

steven9909:


invertedai/helpers/waypoints.py
```diff
@@ -519,20 +533,39 @@ def generate_lane_ids_from_lanelet_map(
...
             return []
-        route = [starting_lanelet.id]
+        route_lane_ids = [starting_lanelet.id]
         total_distance = 0
         while ending_lanelets and total_distance <= (min_distance if min_distance is not None else 0):
-            ending_lanelet = rng.choice(ending_lanelets)
-            possible_route = routing_graph.getRoute(starting_lanelet, ending_lanelet, withLaneChanges=True)
-            if not possible_route:
-                continue
+            routes = [(routing_graph.getRoute(starting_lanelet, ending_lanelet, withLaneChanges=True), ending_lanelet) for ending_lanelet in ending_lanelets]
```
steven9909, line 540: This is going to slow down the waypoint calculation even more, will this be a problem?

KieranRatcliffeInvertedAI: That depends on _how_ much slower it makes the script. Can you do a quick analysis with and without this change? Also a comparison to no waypoint manager whatsoever as a control would be extra helpful.

steven9909: Hmm i think its kind of hard to do a comparison. It all depends on how fast the API server is at this time, and also a function of how many agents, how big the map is, etc.. In the future we might want to integrate something into our CI/CD pipeline with our models running locally, and a whole bunch of experiments with varying map size, agent count? I am just concerned that if we come up with a number right now, its not an accurate representation of how slow it is vs not.

steven9909: https://docs.google.com/spreadsheets/d/1Bnl1XHzlbHL2DlzPpsE2dTAzLttKZhnm6Grdh0a69gE/edit?usp=sharing

google docs detailing the runtime of each methods.

KieranRatcliffeInvertedAI: Thank you for the numbers. The increase in execution time between latest and develop is low enough that I would approve this PR but I'll make a note in the documentation so users are aware. We should look into improving this algorithm over time but is good enough for now.

PR #277, opened 2026-01-15
Update documentation and example files for easier onboarding.

KieranRatcliffeInvertedAI:


examples/scenario_replay_ego_takeover_example.py
```diff
@@ -19,49 +19,72 @@ def main(args):
...
+    if args.num_agents > 0:
+        regions = iai.get_regions_default(
+            location = location,
+            agent_count_dict = {AgentType.car: args.num_agents},
+            area_shape = (int(args.width/2),int(args.height/2)),
+            map_center = scenario_center
+        )
+    else:
+        regions = [
+            iai.large.common.Region.create_square_region(
```
steven9909, line 48: Should we add iai.large as a module in __init__.py? Currently, there is no type hinting for this call

invertedai/logs/logger.py
```diff
@@ -145,7 +145,7 @@ def scenario_log_from_debug_log(
             drive_model_version=all_drive_responses[-1]["model_version"],
             light_recurrent_states=all_drive_responses[-1]["light_recurrent_states"],
             recurrent_states=[RecurrentState.fromval(rec_state) for rec_state in all_drive_responses[-1]["recurrent_states"]],
-            waypoints={str(i):[prop.waypoint] for i, prop in enumerate(agent_properties)},
+            waypoints={str(i):prop.waypoints for i, prop in enumerate(agent_properties)},
```
steven9909, line 148: I think the ScenarioLog class is missing the waypoints attribute

KieranRatcliffeInvertedAI: By design, the ScenarioLog data type has a property of waypoints_per_frame which is a dictionary of vehicle ID's mapped to a list of waypoints.

examples/scenario_replay_ego_takeover_example.py
KieranRatcliffeInvertedAI: This would make sense as our other main functions are included in the package like normal drive and initialize.

adscib (approved):

PR #278, opened 2026-01-27
SimulationManager

atao-inverted:
Generate agents from iai.initialize calls

- Users can optionally pass keys per agent requested (whether pre-existing or sampled agents)

- Default behaviour: give all agents a UUID keys to store in agent_dict{}

Provide getters and setters for the AgentData so the user can manipulate them at will

Add Waypoints as a separate field into the AgentData (thus allowing us to not have to rely on InitializeResponse.AgentProperties)

Calls iai.large_drive and iai.large_initialize when you call AgentDataManager.drive and AgentDataManager.Initialize and packs/unpacks agent states/properties/recurrent states as necessary. 

- Allow users to pass kwargs for all the miscellaneous parameters to these functions

Allow the AgentDataManager to take in different class Config and therefore create new Configs for classes (such as a LoggerConfig) as necessary to initialize them within AgentDataManager

Integrate these different classes (like the waypoint manager, logger) into the AgentDataManager class before and after the iai.initialize/iai.drive calls

Conduct data validation for all data and configuration combinations with appropriate error messages if data is formatted incorrectly


examples/keyed_minimal_example.py
```diff
@@ -21,8 +21,9 @@
 location_info_response = iai.location_info(location=location)
 
 agents = KeyedAgents(num_agents=num_agents_to_add)
```
atao-inverted: initialize agents using KeyedAgents class

invertedai/api/drive.py
```diff
@@ -253,7 +276,13 @@ def _tolist(input_data: List):
...
                 else None
             )
-
+            if keyed_agents is not None:
+                keyed_agents.pack(
+                    agent_ids = agent_ids,
+                    states = response.agent_states,
+                    properties = keyed_properties,
+                    recurrent_states = response.recurrent_states
+                )
```
atao-inverted: Once we get our response we must pack the response.states, response.properties, repsonse.recurrent_states back into the dictionary from our KeyedAgents class

examples/keyed_large_map_example.py
KieranRatcliffeInvertedAI, line 1: I think making 3 additional example files is a tad excessive and will muddy our examples directory.

KieranRatcliffeInvertedAI: I think one example file should suffice at most. It may be easier to add this to an existing file instead.

examples/keyed_large_map_example.py
```diff
@@ -0,0 +1,178 @@
...
+
+        total_num_agents = len(response.agent_states)
+        print(f"Number of agents in simulation: {total_num_agents}")
+        agents = KeyedAgents(num_agents=total_num_agents)
+        agents.set_states(response.agent_states)
+        agents.set_properties(response.agent_properties)
+        agents.set_recurrent_states(response.recurrent_states)
+        print(f"Begin stepping through simulation.")
+        for _ in tqdm(range(args.sim_length)):
+            response = iai.large_drive(
```
KieranRatcliffeInvertedAI: The functionality of the keyed agents are meant to replicate the same assumptions as the current SDK functionality. In this case, it means the mutability of parameters needs to remain the same (i.e. values should be passed by value, not be reference). This means that you will need to update the DriveResponse and InitializeResponse classes to return a KeyedAgents object with the update AgentData with the same keys.

KieranRatcliffeInvertedAI: My vision is that the change to a script using keyed_agents vs what currently exists will only change by a few lines. Those lines being where the data is retrieved from the response object (e.g. for subsequent drive() calls or getting the data for logging/visualization).

invertedai/large/drive.py
KieranRatcliffeInvertedAI, line 1: large_initialize should be updated as well.

invertedai/keyed_agent.py
```diff
@@ -0,0 +1,134 @@
...
+    Class for managing keyed agents with an internal dictionary structure to manage AgentData by AgentID
+    """
+    agents_dict: AgentDict
+    def __init__(self, num_agents: int = 0, **data):
+        agents_dict = {}
+        for i in range(num_agents):
+            aid = f"agent_{i}"
+            agents_dict[aid] = AgentData(
+                properties=get_default_agent_properties({AgentType.car: 1})[0]
+            )
```
KieranRatcliffeInvertedAI: This can be reduced to a single line by passing `num_agents` to `get_default_agent_properties()`.

KieranRatcliffeInvertedAI (changes requested):

invertedai/api/drive.py
```diff
@@ -86,7 +87,7 @@ def serialize_drive_request_parameters(
 @validate_call
 def drive(
     location: str,
-    agent_states: List[AgentState],
+    agent_states: Optional[List[AgentState]] = None,
```
KieranRatcliffeInvertedAI: Since agent_states is possibly Optional now, there should be a parameter validation step checking to make sure key_agents and agent_states are both not None.

KieranRatcliffeInvertedAI: This should be repeated anywhere agent_states has become Optional due to the existence of keyed_agents.

invertedai/logs/logger.py
```diff
@@ -295,31 +299,49 @@ def export_to_file(
...
+                    individual_suggestions_dict[str(i)] = {
+                        "suggestion_strength": 0.8, #Default value
+                        "states":[{
+                            "center": {
+                                "x": wp.x,
+                                "y": wp.y
+                            }
+                        }]
+                    }
+            if scenario_log.waypoints_per_frame is not None:
```
atao-inverted: logging generated waypoints from the WaypointManager

invertedai/helpers/waypoints.py
```diff
@@ -52,7 +56,7 @@ class WaypointManager:
     
     def __init__(
         self,
-        location_info_response: LocationResponse,
+        lanelet_map: Optional[lanelet2.core.LaneletMapLayers] = None, # only OSM map
```
KieranRatcliffeInvertedAI: It is a good change to only include the lanelet2 map instead of the whole LocationInfoResponse but this should be taken one step further and added to the config dataclass instead.

KieranRatcliffeInvertedAI: I see that the lanelet2 map is indeed added to the config dataclass. I assume including this in the init() function is a mistake? Please change it and ensure *every* reference to the WaypointManager is updated to reflect this change.

KieranRatcliffeInvertedAI (changes requested):

examples/keyed_minimal_example.py
```diff
@@ -0,0 +1,45 @@
...
+scene_plotter_cfg = ScenePlotterConfig(location=location, location_info_response=location_info_response)
+waypoint_cfg = WaypointManagerConfig(lanelet_map = location_info_response.get_lanelet_map())
+log_cfg = LogWriterConfig(log_path="keyed_minimal_example_log.json",location=location, location_info_response=location_info_response)
+agents = SimulationManager(num_agents=num_agents_to_add, scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg)
+print("initialized agents with ids ", agents.get_agent_ids())
+response = agents.initialize(location=location)
+rendered_static_map = location_info_response.birdview_image.decode()
+
+print("Begin stepping through simulation.")
+for step in range(150):
```
KieranRatcliffeInvertedAI: Hard-coded values when applicable, should be declared as variables in SCREAMING_SNAKE_CASE and placed somewhere obvious like the top of the script. The other hard coded variables have not been declared in SCREAMING_SNAKE_CASE so this would be a good opportunity to write them as such.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+AgentID = str                
+class AgentData(BaseModel):
+    """
+    Container for all agent data
+    """
+    state: Optional[AgentState] = None
+    properties: Optional[AgentProperties] = None
+    recurrent: Optional[RecurrentState] = None
+    # waypoint: Optional[Point] = None
+    # waypoints: Optional[List[Point]] = None
```
KieranRatcliffeInvertedAI: In the client-facing SDK there shouldn't be any commented-out fields like this.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+
+class SimulationManager: 
+    """
+    Class for managing keyed agents with an internal dictionary structure to manage AgentData by AgentID
+    """
+    def __init__(
+            self,
+            scene_plotter_cfg: Optional[ScenePlotterConfig] = None,
+            agents_dict: Optional[AgentDict] = None,
+            num_agents: int = 0,
```
KieranRatcliffeInvertedAI: What is the purpose of this field when agents_dict exists? Please add a docstring like we have for other classes in our SDK explaining what this tool does, what the parameters mean, and what each field does. You will likely need to also document each cfg file.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+            if log_writer_cfg:
+                self.log_writer = LogWriter()    
+
+
+    def add_agent(
+        self,
+        agent_id: str,
+        agent: AgentData,
+        overwrite: bool = False,
+    ):
```
KieranRatcliffeInvertedAI: This function should be extended to write a list of agents instead which also covers the case of a single agent therefore is more future-proof.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        self.agents_dict[agent_id] = agent
+
+    def remove_agent(self, agent_id: str) -> AgentData:
+        """
+        Remove an agent from the dictionary
+        Returns the removed AgentData
+        """
+        if agent_id not in self.agents_dict:
+            raise KeyError(f"Agent '{agent_id}' does not exist")
+        return self.agents_dict.pop(agent_id)
```
KieranRatcliffeInvertedAI: Same comment here, make this a list of agents to be removed in case they want to remove many at once.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+    def unpack(
+        self
+    ) -> Tuple[
+        List[str],
+        List[AgentState],
+        List[AgentProperties],
+        List[RecurrentState],
+    ]:
+        """
+        Convert AgentDict into List[AgentID] + 3 lists for API
```
KieranRatcliffeInvertedAI: If this function is intended to be used exclusively internally to the class, it should be named with a leading "_" as convention.

KieranRatcliffeInvertedAI: If this function is meant to be internal only, remove the docstring. If it is meant to be an external function available for developers to call, the docstring should be updated to be more clear. Suggested edit "Converts the internal agents_dict data into a format that are accepted by the API".

KieranRatcliffeInvertedAI: If this is meant to be an external function, no change required.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+    def pack(
+        self,
+        agent_ids: List[str],
+        states: List[AgentState],
+        properties: Optional[List[AgentProperties]],
+        recurrent_states: Optional[List[RecurrentState]],
+    ):
+        """"
+        Reconstruct AgentDict from API response
+        """
```
KieranRatcliffeInvertedAI: Same comments as with the unpack function regarding naming and docstring changes.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        self.agents_dict = {
+            aid: AgentData(
+                state=states[i],
+                properties=properties[i] if properties else None,
+                recurrent=recurrent_states[i] if recurrent_states else None,
+            )
+            for i, aid in enumerate(agent_ids)
+        }
+    
+    def initialize(self, location: str, **kwargs)->InitializeResponse:
```
KieranRatcliffeInvertedAI: Please place the arguments on separate lines like the other functions defined in this class.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+            self.log_writer.initialize(
+                location=self.log_writer_cfg.location,
+                location_info_response=self.log_writer_cfg.location_info_response,
+                init_response=response,
+                waypoints=waypoints 
+            )
+
+        return response
+    
+    def drive(self, location: str, **kwargs)-> DriveResponse:
```
KieranRatcliffeInvertedAI: Please place the arguments on separate lines like the other functions defined in this class.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        regions = iai.get_regions_default(
+            location = location,
+            agent_count_dict = {AgentType.car: len(self.agents_dict)},
+        )
+        response = iai.large_initialize( 
+            location=location,
+            regions=regions,
+            agent_properties=properties,
+            **kwargs
+        )
```
KieranRatcliffeInvertedAI: The purpose of calling unpack() here is for the case of pre-existing agents correct? If that's the case, the way large_initialize() is formatted is incorrect. Use the agent_states field similar to how you've used the agent_properties field for pre-existing agents.

len(self.agents_dict) gives you the number of already existing agents. You need to create a method for the user to initialize a number of new agents.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        num_returned = len(response.agent_states)
+        if num_returned > len(agent_ids):
+            # API returned more agents than we seeded 
+            agent_ids = agent_ids + [str(uuid.uuid4()) for _ in range(num_returned - len(agent_ids))]
+        elif num_returned < len(agent_ids):
+            agent_ids = agent_ids[:num_returned]
+        print("agent states", len(response.agent_states))
```
KieranRatcliffeInvertedAI: Do not include a print statement in the client facing SDK code (except for example scripts). If anything make this a debug statement into the logging tool here: https://github.com/inverted-ai/invertedai/blob/develop/invertedai/logs/diagnostics.py

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+            regions=regions,
+            agent_properties=properties,
+            **kwargs
+        )
+        num_returned = len(response.agent_states)
+        if num_returned > len(agent_ids):
+            # API returned more agents than we seeded 
+            agent_ids = agent_ids + [str(uuid.uuid4()) for _ in range(num_returned - len(agent_ids))]
+        elif num_returned < len(agent_ids):
+            agent_ids = agent_ids[:num_returned]
```
KieranRatcliffeInvertedAI: This seems rather clunky to me. It should never return MORE agents than requested. If you are using pre-existing agents correctly in the large_initialize() call, they will be at the beginning of the list of response.agent_states and it will contain the exact number you gave it. Therefore you should be able to get the exact amount of new agents that can then be packed into the agents_dict object.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+            self.scene_plotter.initialize_recording(
+                agent_states=response.agent_states,
+                agent_properties=response.agent_properties,
+            )
+        self.pack(
+            agent_ids=agent_ids,
+            states=response.agent_states,
+            properties=properties, ## # consider isolating later
+            recurrent_states=response.recurrent_states,
+        )
```
KieranRatcliffeInvertedAI: For readability sake, this packing should be done immediately after the new agent data is acquired unless this data will be modified in some way before it is packed. In this case, the packing should be done after the WaypointManager step.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+                response.agent_states,
+                traffic_light_states=response.traffic_lights_states,
+                agent_properties=properties,
+            )
+        self.pack(
+            agent_ids=agent_ids,
+            states=response.agent_states,
+            properties=properties,
+            recurrent_states=response.recurrent_states,
+        )
```
KieranRatcliffeInvertedAI: For readability sake, this packing should be done immediately after the new agent data is acquired unless this data will be modified in some way before it is packed. In this case, the packing should be done after the WaypointManager step.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+                    for aid, data in self.agents_dict.items()
+                    if data.properties is not None and data.properties.waypoints is not None
+                }
+            self.log_writer.drive(
+                drive_response=response,
+                waypoints=waypoints 
+            )
+        return response
+    
+    def animate_scene(self, **kwargs) -> FuncAnimation:
```
KieranRatcliffeInvertedAI: I would change the name of this function to be something more specific like "visualize_data".

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        """
+        wrapper around scenePlotter animate...
+        """
```
KieranRatcliffeInvertedAI, line 382: Docstrings should describe what the function does functionally at a high level, not how the function is implemented. As it is written currently, it is not future-proof and will need to be changed whenever we change the implementation of the function even if the overall functionality does not change.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+        if self.scene_plotter is None:
+            raise ValueError("ScenePlotter not initialized, failed to animate scene")
+        return self.scene_plotter.animate_scene(**kwargs)
```
KieranRatcliffeInvertedAI: This function should not return anything. It writes a GIF to the given path or throws an error.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+                init_response=response,
+                waypoints=waypoints 
+            )
+
+        return response
+    
+    def drive(self, location: str, **kwargs)-> DriveResponse:
+        """
+        Wrapper around iai.large_drive
+        """
```
KieranRatcliffeInvertedAI, line 302: Docstrings should describe what the function does functionally at a high level, not how the function is implemented. As it is written currently, it is not future-proof and will need to be changed whenever we change the implementation of the function even if the overall functionality does not change.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,292 @@
...
+                properties=properties[i] if properties else None,
+                recurrent=recurrent_states[i] if recurrent_states else None,
+            )
+            for i, aid in enumerate(agent_ids)
+        }
+    
+    def initialize(self, location: str, **kwargs)->InitializeResponse:
+        """
+        Wrapper around iai.large_initialize
+        """
```
KieranRatcliffeInvertedAI, line 205: Docstrings should describe what the function does functionally at a high level, not how the function is implemented. As it is written currently, it is not future-proof and will need to be changed whenever we change the implementation of the function even if the overall functionality does not change.
React

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI: This comment has not been addressed.

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI: This comment has not been addressed.

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI: This is better than it was before however I would say since this function primarily just passes its parameters on to ScenePlotter, any arguments that are not explicitly defined here should instead be covered in a single sentence telling the reader to check the documentation there. That will make this docstring more future-proof and we won't have to update multiple docstrings if the implementation changes.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+            self.agents_dict = agents_dict
+            self.waypoint_manager: Optional[WaypointManager] = None
+            if waypoint_cfg:
+                self.waypoint_manager = WaypointManager(cfg=waypoint_cfg)
+            self.log_writer = None
+            self.log_writer_cfg = log_writer_cfg
+            if log_writer_cfg:
+                self.log_writer = LogWriter()   
+    
+    def insert_agents_from_lists(
```
KieranRatcliffeInvertedAI: Change this functions name to simply "insert_agents"

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+            if data.recurrent is not None:
+                recurrent_states.append(data.recurrent)
+        return agent_ids, states, properties, recurrent_states
+    
+    def _pack(
+        self,
+        agent_ids: List[str],
+        states: List[AgentState],
+        properties: Optional[List[AgentProperties]],
+        recurrent_states: Optional[List[RecurrentState]],
```
KieranRatcliffeInvertedAI: These types should not be Optional[List[Type]] but should instead be List[Optional[Type]] because it is possible for some agents to be pre-defined without properties or recurrent states and some not. However this is a bit clunky for a few reasons including data validation.

A better solution is define another data type as follows:

```
@dataclass
class APIAgentData:
    agent_id: str
    state: AgentState
    properties: Optional[AgentProperties] = None
    recurrent_state: Optional[RecurrentState] = None
```

Then the `_pack()` and `_unpack()` functions can use this data type instead of Tuples or defining clunkier expected types like I outlined above.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+import uuid
+
+AgentID = str                
+class AgentData(BaseModel):
+    """
+    Container for all agent data
+    """
+    state: Optional[AgentState] = None
+    properties: Optional[AgentProperties] = None
+    recurrent: Optional[RecurrentState] = None
```
KieranRatcliffeInvertedAI: I know we've been a tad inconsistent with the dataclass decorator vs BaseModel inheritance but we should start being consistent here, especially in this class which is bound to be much more important.

We should use the @dataclass decorator when the class simply holds data and use BaseModel when it needs to perform more complex functions with the data inside it (e.g. specific data validation beyond checking if it's the correct type like checking if lists are the same length).

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI: Oh apologies this class more or less already exists. Please replace what I outlined above with the AgentData class COMBINED WITH AgentDict to simplify the data types in the `_pack()` and `_unpack()` functions.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+        for i, agent_id in enumerate(self.agents_dict.keys()):
+            self.agents_dict[agent_id].recurrent = recurrent_states[i]
+
+def build_agent_dict_from_lists(
+    states: List[AgentState],
+    properties: List[AgentProperties],
+    recurrent_states: Optional[List[Optional[RecurrentState]]] = None,
+    existing_agents_dict: Optional[AgentDict] = None,
+    overwrite: bool = False,
+) -> AgentDict:
```
KieranRatcliffeInvertedAI: Is this function used anywhere? If not, please let me know if there's a good reason to keep it here otherwise it should be removed.

examples/keyed_minimal_example.py
```diff
@@ -0,0 +1,45 @@
...
+api_key = os.environ.get("IAI_API_KEY", None)
+if api_key is None:
+    iai.add_apikey("<INSERT_KEY_HERE>")
+
+print("Begin initialization.")
+location_info_response = iai.location_info(location=LOCATION, include_map_source=True)
+scene_plotter_cfg = ScenePlotterConfig(location=LOCATION, location_info_response=location_info_response)
+waypoint_cfg = WaypointManagerConfig(lanelet_map = location_info_response.get_lanelet_map())
+log_cfg = LogWriterConfig(log_path="keyed_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
+agents = SimulationManager(num_agents=NUM_AGENTS, scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg)
```
KieranRatcliffeInvertedAI: I would change this variable name to simulation_manager so it's more clear what it is.

examples/keyed_minimal_example.py
KieranRatcliffeInvertedAI, line 1: I think the simulation manager class you've written is so useful and effective, please replace the existing `minimal_example.py` with this so users can see how easy it is to get started.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+            for i, aid in enumerate(agent_ids)
+        }
+    
+    def initialize(
+            self, 
+            location: str, 
+            regions: Optional[List[Region]] =None, 
+            num_new_agents: Optional[int] = None,
+            **kwargs
+        )->InitializeResponse:
```
KieranRatcliffeInvertedAI: Please fix the indentation here.

examples/keyed_minimal_example.py
KieranRatcliffeInvertedAI, line 1: I think you should create another example script that demonstrates cosimulation with the SimulationManager as that is one of the main functionalities of our SDK.

invertedai/simulation_manager.py
atao-inverted: Unfortunately the conversion from AgentData -> list(AgentStates), list(AgentProperties), list(RecurrentStates) must be done to call the API. 

If `pack` and `unpack` return an AgentData object to reduce the use of tuples, it sort of makes these two functions redundant ? Since our API only uses lists

examples/keyed_minimal_example.py
atao-inverted: Sounds good. 

Just wanted to mention that this example script currently has the Logwriter, Sceneplotter, and WaypointManager enabled, so I'm not sure if it is strictly a 'minimal' example

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+        self.agents_dict = {
+            aid: AgentData(
+                state=states[i],
+                properties=properties[i] if properties else None,
+                recurrent=recurrent_states[i] if recurrent_states else None,
+            )
+            for i, aid in enumerate(agent_ids)
+        }
+    
+    def initialize(
```
atao-inverted: data validation from unpack

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+            self, 
+            location: str, 
+            regions: Optional[List[Region]] =None, 
+            num_new_agents: Optional[int] = None,
```
atao-inverted: To summarize from our chat:

Regions -> not optional

num_new_agents -> remove

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+            **kwargs
```
atao-inverted: To summarize from our chat:

Add Optional[AgentDict] to insert external Agents with their keys. 

Then use insert_agent with overwrite flag turned on

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,439 @@
...
+        if missing:
+            raise KeyError(f"Agents do not exist: {missing}")
+
+        removed = {}
+        for aid in agent_ids:
+            if aid not in self.agents_dict:
+                raise KeyError(f"Agent '{aid}' does not exist")
+            return self.agents_dict.pop(aid)
+
+        return removed
```
atao-inverted: To summarize from our chat:

should not return anything

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,436 @@
...
+        return response
+    
+    def drive(
+        self, 
+        location: str, 
+        **kwargs
+    )-> DriveResponse:
+        """
+        Advance the simulation by one timestep using the current agents in self.agent_dict
+
```
atao-inverted: From our chat:

Ability to overwrite with external agent data with Overwrite= TRUE inserting into our agent_dict at each step

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,417 @@
...
+        for i, agent_id in enumerate(self.agents_dict.keys()):
+            self.agents_dict[agent_id].recurrent = recurrent_states[i]
+
+def build_agent_dict_from_lists(
+    states: List[AgentState],
+    properties: List[AgentProperties],
+    recurrent_states: Optional[List[Optional[RecurrentState]]] = None,
+    existing_agents_dict: Optional[SimulationAgentData] = None,
+    overwrite: bool = False,
+) -> SimulationAgentData:
```
KieranRatcliffeInvertedAI: I still do not have an answer to why this function is here? Is there a good reason? If not, it should be delete.

atao-inverted: Fair point. I will delete it for now and only add it later when necesary

examples/keyed_minimal_example.py
KieranRatcliffeInvertedAI, line 1: Rename this function to `simulation_manager_minimal_example.py`

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,357 @@
...
+        regions: List[Region],
+        external_agent_data: Optional[SimulationAgentDict] = None, # optional param for passing in external agent data in the form of a SimulationAgentDict. Overwrite = True 
+        **kwargs
+    ) -> InitializeResponse:
+        """
+        Wrapper around iai.large_initialize
+        
+        Please see iai.large_initialize for documentation on kwargs
+
+        """
```
KieranRatcliffeInvertedAI, line 205: Please include documentation for any arguments that are explicitly included in this function and why they are not just lumped into kwargs. For instance, why is `location` an explicit parameter and not included in kwargs?

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,357 @@
...
+        Advance the simulation by one timestep using the current agents in self.agent_dict
+
+        This method:
+        - updated the self.agent_dict with results from DRIVE
+        - Applies waypoint-based modifications if configured
+        - Records visualization and logging outputs if configured
+
+        Returns:
+            DriveResponse
+        """
```
KieranRatcliffeInvertedAI, line 302: Please include documentation for any arguments that are explicitly included in this function and why they are not just lumped into kwargs. For instance, why is location an explicit parameter and not included in kwargs?

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,357 @@
...
+        kwargs.pop("recurrent_states", None)
```
KieranRatcliffeInvertedAI: Why is recurrent_states popped here? In terms of users including keyword arguments that are already taken care of by unpack(). we should pick a lane in regards to data validation. An error will be thrown automatically if two arguments with the same keyword are passed so we may not necessarily need to do this but perhaps we should do a data validation step for checking each states, properties, and recurrent states.

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI, line 302: Please include a note here that states, properties, and recurrent states do not need to be included in kwargs but for the rest of the possible parameters, please checkout large_drive and include a link.

invertedai/simulation_manager.py
KieranRatcliffeInvertedAI, line 205: Please include a note here that states, properties, and recurrent states do not need to be included in kwargs but for the rest of the possible parameters, please checkout large_initialize and include a link.

KieranRatcliffeInvertedAI: My apologies by link I meant something more like this in large_drive for example: https://github.com/inverted-ai/invertedai/blob/develop/invertedai/large/drive.py#L83

examples/simulation_manager_external_agent_example.py
```diff
@@ -0,0 +1,61 @@
...
+    iai.add_apikey("<INSERT_KEY_HERE>")
+
+print("Begin initialization.")
+log_reader.initialize()
+location_info_response = iai.location_info(location=LOCATION, include_map_source=True)
+scene_plotter_cfg = ScenePlotterConfig(location=LOCATION, location_info_response=location_info_response)
+waypoint_cfg = WaypointManagerConfig(lanelet_map = location_info_response.get_lanelet_map())
+log_cfg = LogWriterConfig(log_path="keyed_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
+simulation_manager = SimulationManager(scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg)
+regions = iai.get_regions_default(agent_count_dict = {AgentType.car: NUM_AGENTS}, location = LOCATION, map_center=tuple([location_info_response.map_center.x, location_info_response.map_center.y]))
```
KieranRatcliffeInvertedAI: Please follow the style guide for formatting this line.

examples/simulation_manager_external_agent_example.py
```diff
@@ -0,0 +1,61 @@
...
+external_agent_data = {
+    "new_agent": AgentData(
+        state=log_reader.agent_states[0], # add in the state of the external agent
+        properties=log_reader.agent_properties[0], # add in the properties of the external agent
+        recurrent=None # no recurrent state necessary for initialization
+    )
+}
+response = simulation_manager.initialize(location=LOCATION, regions=regions, external_agent_data=external_agent_data)
```
KieranRatcliffeInvertedAI: Please follow the style guide for formatting this line and all other function calls with multiple input parameters.

examples/simulation_manager_external_agent_example.py
```diff
@@ -0,0 +1,90 @@
...
+    log_writer_cfg=log_cfg
+)
+regions = iai.get_regions_default(
+    agent_count_dict = {AgentType.car: NUM_AGENTS}, 
+    location = LOCATION, 
+    map_center=tuple([location_info_response.map_center.x, location_info_response.map_center.y])
+)
+external_agent_data = {
+    # please use the Scenario Builder tool to check the validity of the agent states
+    "new": AgentData(
```
KieranRatcliffeInvertedAI: Please provide a better agent ID than "new". Something that is more in-line with what any arbitrary agent, whether pre-defined or sampled, would have.

examples/simulation_manager_external_agent_example.py
```diff
@@ -0,0 +1,90 @@
...
+    waypoint_cfg=waypoint_cfg, 
+    log_writer_cfg=log_cfg
+)
+regions = iai.get_regions_default(
+    agent_count_dict = {AgentType.car: NUM_AGENTS}, 
+    location = LOCATION, 
+    map_center=tuple([location_info_response.map_center.x, location_info_response.map_center.y])
+)
+external_agent_data = {
+    # please use the Scenario Builder tool to check the validity of the agent states
```
KieranRatcliffeInvertedAI: Suggested comment change: "You can use the Scenario Builder tool available on the Inverted AI website to validate an agent's state and properties."

examples/simulation_manager_external_agent_example.py
```diff
@@ -0,0 +1,90 @@
...
+)
+print("initialized agents with ids ", simulation_manager.get_agent_ids())
+rendered_static_map = location_info_response.birdview_image.decode()
+
+print("Begin stepping through simulation.")
+for step in range(SIM_LENGTH):
+    response = simulation_manager.drive(
+        location=LOCATION, 
+        light_recurrent_states=response.light_recurrent_states
+    )
```
KieranRatcliffeInvertedAI: This is meant to be a co-simulation example. Part of co-simulation is that the agents are also driven externally. Please demonstrate that here.

examples/simulation_manager_external_agent_example.py
KieranRatcliffeInvertedAI, line 1: Why does this file exist if simulation_manager_cosimulation_minimal_example.py exists?

examples/simulation_manager_cosimulation_minimal_example.py
```diff
@@ -0,0 +1,108 @@
...
+        recurrent_states=ego_response.recurrent_states, 
+    )
+    external_agent_data = {
+            ego_agent_ids[i]: AgentData(
+                state=ego_response.agent_states[i],
+                properties=ego_props[i],
+                recurrent=None,  # recurrent is always zeroed for external agents in SimulationManager.drive()
+            )
+            for i in range(NUM_EGO_AGENTS)
+        }
```
KieranRatcliffeInvertedAI: Fix the indentation here.

examples/simulation_manager_cosimulation_minimal_example.py
```diff
@@ -0,0 +1,108 @@
...
+print("Simulation finished, save visualization.")
+
+fig, ax = plt.subplots(constrained_layout=True, figsize=(10, 10))
+simulation_manager.visualize_data(
+    output_name="simulation_manager_cosimulation_example.gif",
+    ax=ax,
+    direction_vec=False,
+    velocity_vec=False,
+    plot_frame_number=True,
+    numbers = list(range(NUM_AGENTS))
```
KieranRatcliffeInvertedAI: Change this to be all agents. I think it will better demonstrate how to use the visualization tool.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+        Note:
+        - agent_states, agent_properties, and recurrent_states should not be
+          provided in kwargs. These values are automatically derived from the
+          internal agent dictionary and managed by this wrapper
+        - For all other supported parameters, please refer to the documentation for :func:`large_initialize`        
+        """
+        temp_dict = dict(self.agents_dict)
+        if external_agent_data:
+            temp_dict.update(external_agent_data)
+        agent_ids, states, properties, recurrent_states = self._unpack(temp_dict)
```
KieranRatcliffeInvertedAI: What about pre-existing agents already in the agent_dict? They should be unpacked and included too. However, I like the way unpack is designed such that it either uses an agent_dict parameter OR the member agent_dict and not both which keeps things clean. So you will have to call unpack twice in the case external agents are included.

KieranRatcliffeInvertedAI: There must be a way to write this in a more pythonic way that does not rely on a temporary variable. Doing so will make the code more readable and maintainable.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+        response = iai.large_initialize( 
+            regions=regions,
+            agent_properties=properties,
+            agent_states=states,
+            return_exact_agents=True,
+            **kwargs
+        )
+
+        num_new_agents = len(response.agent_states) - original_agent_count
+        new_ids = [str(uuid.uuid4())[:8] for _ in range(num_new_agents)]
```
KieranRatcliffeInvertedAI: Is there a reason to only use the first 8 digits that supersedes the benefit of the reduction of the risk of ID overlap that all digits from a UUID provides?

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+                agent_properties=response.agent_properties,
+            )
+        if self.log_writer is not None:
+            if self.waypoint_manager is not None:
+                waypoints = {
+                    aid: new_properties[i].waypoints
+                    for i, aid in enumerate(all_agent_ids)
+                    if new_properties[i] is not None and new_properties[i].waypoints is not None
+                }
+            self.log_writer.initialize(  # external agents must be included if exist -> make stateless
```
KieranRatcliffeInvertedAI: Remove this comment because it is a note to the developer and not useful to the user. Please remove all other such comments.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+
+        original_agent_count = len(agent_ids)
+
+        response = iai.large_initialize( 
+            regions=regions,
+            agent_properties=properties,
+            agent_states=states,
+            return_exact_agents=True,
+            **kwargs
+        )
```
KieranRatcliffeInvertedAI, line 227: What data validation steps can be done on the data lists here before they are sent to large_initialize()? Are there any validation steps that are not performed by large_initialize() that should be done here? If not, then disregard this comment.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+          provided in kwargs. These values are automatically derived from the
+          internal agent dictionary and managed by this wrapper
+        - For all other supported parameters, please refer to the documentation for :func:`large_drive`
+        """
+        temp_dict = dict(self.agents_dict)
+        _, _, _, internal_recurrent_states = self._unpack(self.agents_dict)
+        internal_recur_size = None
+        for r in internal_recurrent_states:
+            internal_recur_size = len(r.packed)
+            break
```
KieranRatcliffeInvertedAI:
```
if len(internal_recurrent_states) > 0: internal_recur_size = len(internal_recurrent_states[0].packed)
else: Use default
```

Use this to save yourself 2 lines and be more readable.

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,410 @@
...
+                    state=data.state,
+                    properties=data.properties,
+                    recurrent=RecurrentState(packed=[0.0] * internal_recur_size),
+                )
+                for aid, data in external_agent_data.items()
+            }
+            temp_dict.update(zeroed_external)
+        external_ids = set(external_agent_data.keys()) if external_agent_data else set()
+
+        agent_ids, states, properties, recurrent_states = self._unpack(temp_dict)
```
KieranRatcliffeInvertedAI: This whole block can be made cleaner. 

1. Unpack internal agent dict.
2. Get recurrent state size or use default
3. Unpack external agent dict if provided
4. Format external recurrent states using recurrent state size
5. Combine internal and external data lists as necessary.

KieranRatcliffeInvertedAI: Recycle code you've already written here as necessary but please eliminate the reliance on temporary variables.

KieranRatcliffeInvertedAI (changes requested):

invertedai/simulation_manager.py
```diff
@@ -0,0 +1,425 @@
...
+        Note:
+        - agent_states, agent_properties, and recurrent_states should not be
+          provided in kwargs. These values are automatically derived from the
+          internal agent dictionary and managed by this wrapper
+        - For all other supported parameters, please refer to the documentation for :func:`large_initialize`        
+        """
+        if external_agent_data:
+            overlap = set(self.agents_dict.keys()) & set(external_agent_data.keys())
+            if overlap:
+                raise ValueError(f"External agent IDs conflict with internal agents: {overlap}")
```
KieranRatcliffeInvertedAI, line 209: Good addition.

KieranRatcliffeInvertedAI (approved):
Looks good to merge!

KieranRatcliffeInvertedAI (approved):

PR #286, opened 2026-03-18
Add from_tensor helper method

steven9909:


atao-inverted (approved):
LGTM

@KieranRatcliffeInvertedAI from my knowledge, this should be able to pass our CI/CD tests but we should first merge the CI/CD PR https://github.com/inverted-ai/invertedai/pull/285 before this one to allow this branch to run against our CI/CD

KieranRatcliffeInvertedAI:
@steven9909 What is this PR?

steven9909:
This PR is to add helper method for converting tensor into Image object. This came up while working on the scenario endpoint since that now expects an Image object in the return response but the model outputs a tensor.

KieranRatcliffeInvertedAI:
> This PR is to add helper method for converting tensor into Image object. This came up while working on the scenario endpoint since that now expects an Image object in the return response but the model outputs a tensor.

@steven9909 understood. Will this then be part of a larger PR to develop an SDK for a scenario endpoint or will that be a separate PR?

KieranRatcliffeInvertedAI (changes requested):
Please merge the develop branch into this one so that we can see if it passes CI/CD.

invertedai/common.py
```diff
@@ -94,6 +91,14 @@ def decode(self):
     def fromval(cls, val):
         return cls(encoded_image=val)
 
+    @classmethod
+    def from_tensor(cls, tensor, encode_format="PNG"):
```
KieranRatcliffeInvertedAI: Please add typing to the inputs and output as well should make that standard from now on. Furthermore, for a function with more than 1 input parameter, please place each parameter on its own line for readability.

steven9909: done

PR #287, opened 2026-03-19
Refactor LogWriter/LogReader to use SimulationAgentDict (take advantage of the keyed agent property)

atao-inverted:


KieranRatcliffeInvertedAI (changes requested):
These changes do not seem to be backwards compatible.

invertedai/logs/logger.py
```diff
@@ -46,9 +48,69 @@ class ScenarioLog(BaseModel):
     light_recurrent_states: Optional[LightRecurrentStates] = None #: As of the most recent time step. Please refer to the documentation of :func:`drive` for further information on this parameter.
     recurrent_states: Optional[List[RecurrentState]] = None #: As of the most recent time step. Please refer to the documentation of :func:`drive` for further information on this parameter.
 
-    waypoints_per_frame: Optional[List[WaypointsDict]] = None # As of the most recent time step. A list of waypoints keyed to agent ID's not including waypoints already passed. These waypoints are not automatically populated into the agent properties.
-    present_indexes: List[List[int]] = None #: List of indexes corresponding to agent_properties for which agents are present at each time step. If None, all agents are present at every time step.
+    @property
```
KieranRatcliffeInvertedAI: The `property` decorator is not necessary for a BaseModel child class.

KieranRatcliffeInvertedAI:
To be clear, the interfaces and exposed data types (such as ScenarioLog) should still allow for backwards compatibility. But the internal functionality can be updated with the new keyed-agent types which should make the code cleaner and more efficient. So here are the specs in more detail:

- Maintain backwards compatibility of externally exposed points (interfaces and data types)
- Add inputs for the new data types and deprecate the old ones to maintain backwards compatibility while also encouraging usage of the new more efficient data types.
- Make a new ScenarioLog format and deprecate the old one to maintain backwards compatibility while also encouraging usage of the new more efficient data type.

invertedai/logs/logger.py
```diff
@@ -46,9 +48,69 @@ class ScenarioLog(BaseModel):
...
+    def present_indexes(self) -> List[List[int]]:
+        """
+        Per timestep list of integer indexes into agent_ids for present agents
+        """
+        all_ids = self.all_agent_keys
+        id_to_idx = {aid: i for i, aid in enumerate(all_ids)}
+        return [
+            [id_to_idx[k] for k in snap.keys() if k in id_to_idx]
+            for snap in self.agent_history
+        ]
```
KieranRatcliffeInvertedAI: This data type can essentially be the new ScenarioLogX data type (whatever we choose to call it). I think for getters, we can have standard terminology:

- Using "at", e.g. `get_properties_at()`,  like you have here can be for accessing data at a specific time step as you've done.
- Using "all", e.g. `get_agent_ids_all()`, means getting all the relevant data throughout the log. This may not apply to all data types. For example getting "all" agent ID's of any agent that has appeared in the log makes sense but getting "all" states does not as agents can come and go.
- Using "current", e.g. 'get_agent_states_current()', means getting the relevant data at the end of the list. For example, `get_agent_states_current()` will get you the agent states from `self.agent_history[-1]`.

I'm tempted to add a parameter to each of these getters (where it makes sense) to specify an ID of an agent of interest but that might make the code super unmaintainable.

invertedai/logs/logger.py
```diff
@@ -314,44 +372,31 @@ def _format_waypoints_json(
...
+        all_keys = scenario_log.all_agent_keys
+        agent_properties = scenario_log.all_agent_properties_map
+
+        # Build waypoints from agent properties map
+        individual_suggestions_dict = {}
+        for key, prop in agent_properties.items():
+            if prop.waypoints:
+                individual_suggestions_dict[key] = _format_waypoints_json(prop.waypoints)
+            elif prop.waypoint:
+                individual_suggestions_dict[key] = {
```
KieranRatcliffeInvertedAI: I know this is leftover from the old code but this block could also make use of the '_format_waypoints_json' function by simply placing the single waypoint into a list.

KieranRatcliffeInvertedAI:
Make sure to also update the example files that demonstrate integration with the Logger tools with the changes you make. This should also serve as a means to test your PR.

Furthermore, please merge in the CI/CD hotfix. It seems to have already worked in its own PR but I would like to see it work here too.

atao-inverted:
@KieranRatcliffeInvertedAI I understand your points. 

So you want to keep our old ScenarioLog and instead create a new ScenarioLogV2 (or other data type name) to maintain backwards compatibility. 

The thing is, I dont see why maintaining strict backwards compatibility of `ScenarioLog` is necessary since ScenarioLog is an _internal_ data structure that LogWriter and LogReader use. In the SDK users never construct a ScenarioLog themselves, Users only interact with LogWriter and LogReader. 

Since ScenarioLog is purely internal -> replacing `agent_states`/`agent_properties`/`present_indexes` fields with `agent_history` (a `SimulationAgentDict`) should be completely fine.

The backwards compatibility preservation points should be in the LogReader/LogWriter interfaces (for example `initialize() `and `drive()`) and these are preserved in my current changes

TLDR: `ScenarioLog` is purely internal. The public API is `LogWriter`/`LogReader` and only they need to be backwards compatible

These current changes above work against all our existing example files under `examples/`


But I hear your point on full backwards compatibility. I will work on making LogReader and LogWriter interfaces backwards compatible

If theres confusion, perhaps a discussion in-person would clear this all up?

KieranRatcliffeInvertedAI:
> @KieranRatcliffeInvertedAI I understand your points.
> 
> So you want to keep our old ScenarioLog and instead create a new ScenarioLogV2 (or other data type name) to maintain backwards compatibility.
> 
> The thing is, I dont see why maintaining strict backwards compatibility of `ScenarioLog` is necessary since ScenarioLog is an _internal_ data structure that LogWriter and LogReader use. In the SDK users never construct a ScenarioLog themselves, Users only interact with LogWriter and LogReader.
> 
> Since ScenarioLog is purely internal -> replacing `agent_states`/`agent_properties`/`present_indexes` fields with `agent_history` (a `SimulationAgentDict`) should be completely fine.
> 
> The backwards compatibility preservation points should be in the LogReader/LogWriter interfaces (for example `initialize() `and `drive()`) and these are preserved in my current changes
> 
> TLDR: `ScenarioLog` is purely internal. The public API is `LogWriter`/`LogReader` and only they need to be backwards compatible
> 
> These current changes above work against all our existing example files under `examples/`
> 
> But I hear your point on full backwards compatibility. I will work on making LogReader and LogWriter interfaces backwards compatible
> 
> If theres confusion, perhaps a discussion in-person would clear this all up?

The issue is that the ScenarioLog format is _mostly_ internal. The `return_scenario_log()` function in LogReader returns the ScenarioLog data type allowing the user to manipulate it and access data fields.

invertedai/logs/logger.py
```diff
@@ -1,26 +1,29 @@
...
     AgentState, 
     LightRecurrentState,
     LightRecurrentStates,
     Point,
     RecurrentState,
+    SimulationAgentDict,
     TrafficLightStatesDict 
 )
-
 class ScenarioLog(BaseModel):
```
KieranRatcliffeInvertedAI, line 28: As for naming structure, I think the new version should be called `ScenarioLog` and the old version be called `ScenarioLogLegacy` with a note about deprecation.

invertedai/logs/logger.py
```diff
@@ -79,7 +82,187 @@ def add_time_step_data(
...
-        self.agent_states.append(current_agent_states)        
+        self.agent_states.append(current_agent_states)   
+
+class ScenarioLogV2(BaseModel):
+    """
+    A log containing simulation information for storage, replay, or an initial state from which a simulation 
+    can be continued. Some data fields contain data for all historic time steps while others contain information
+    for the most recent time step to be used to continue a simulation.
+    """
+    agent_history: List[SimulationAgentDict]
```
KieranRatcliffeInvertedAI: Rename this to `agent_data` instead of `agent_history`. The latter implies this is old and past. The former is more neutral phrasing.

invertedai/logs/logger.py
```diff
@@ -79,7 +82,187 @@ def add_time_step_data(
...
+        """Agent properties at a specific timestep"""
+        return [data.properties for data in self.agent_history[timestep].values()]
+
+    def get_agents_at(self, timestep: int) -> SimulationAgentDict:
+        """ agent dict at a specific timestep"""
+        return self.agent_history[timestep]
+
+    def get_agents_current(self) -> SimulationAgentDict:
+        """agent dict from the latest timestep in agent_history"""
+        return self.agent_history[-1]
```
KieranRatcliffeInvertedAI: I realize my previous instruction for these getters was not consistent with Only Code It Once. I think we don't need the "current" versions at all. What I think can be done instead is to make the "timestep" parameters Optional and if None is provided, it gets the most current data.

KieranRatcliffeInvertedAI: Please implement that change and delete all "current" getters.

invertedai/logs/logger.py
```diff
@@ -79,7 +82,187 @@ def add_time_step_data(
...
+    @model_validator(mode='after')
+    def validate_states_and_present_indexes_init(self):
+        present_indexes = self.get_present_indexes_all()
+        if present_indexes is not None:
+            for t in range(len(self.agent_history)):
+                self._validate_timestep(
+                    current_agent_states=self.get_agent_states_at(t),
+                    current_present_indexes=present_indexes[t]
+                )
+        return self
```
KieranRatcliffeInvertedAI: These functions were necessary when present_indexes was a necessary field to keep track of agents that were coming and going from the simulation. I think these functions are no longer necessary because the SimulationDict inherently keeps track of that information.

invertedai/logs/logger.py
```diff
@@ -79,7 +82,187 @@ def add_time_step_data(
...
+
+    def _validate_timestep(
+        self,
+        current_agent_states: List[AgentState],
+        current_present_indexes: List[int]
+    ):
+        assert min(current_present_indexes) >= 0, "Invalid agent ID's in given list of present indexes."
+        assert len(current_present_indexes) == len(current_agent_states), "Given number of agent states does not match number of present agents."
```
KieranRatcliffeInvertedAI: The `present_indexes` field should be removed during this refactor. The only place it should exist is when converting the new ScenarioLog format to the old ScenarioLog format which can be generated from the List of SimulationDicts.

KieranRatcliffeInvertedAI (changes requested):

invertedai/helpers/simulation_manager.py
KieranRatcliffeInvertedAI, line 1: If I recall, we discussed that LogReader did not need to be integrated into SimulationManager. Can you explain what has changed since then to cause that change in scope?

invertedai/logs/logger.py
```diff
@@ -79,7 +82,187 @@ def add_time_step_data(
...
+            rendering_center=legacy.rendering_center,
+            rendering_fov=legacy.rendering_fov,
+            lights_random_seed=legacy.lights_random_seed,
+            initialize_random_seed=legacy.initialize_random_seed,
+            drive_random_seed=legacy.drive_random_seed,
+            initialize_model_version=legacy.initialize_model_version,
+            drive_model_version=legacy.drive_model_version,
+            light_recurrent_states=legacy.light_recurrent_states,
+            recurrent_states=legacy.recurrent_states,
+        )
```
KieranRatcliffeInvertedAI: We agreed to put this functionality into the legacy ScenarioLog so that when it eventually gets cycled out, it would be a clean delete of the legacy ScenarioLog code. Please place these functions into the legacy version of ScenarioLog.

invertedai/logs/logger.py
atao-inverted: I am slightly worried this could break client's code if the clients have:

`from invertedai.logs.logger import ScenarioLog`

But I will implement the naming fix assuming the above will not break anything

invertedai/logs/logger.py
```diff
@@ -130,10 +313,21 @@ def scenario_log_from_debug_log(
...
-            agent_properties=agent_properties, 
+
+        # Build agent_history from parsed data
+        agent_history = []
+        for t, states_list in enumerate(all_agent_states):
+            agent_dict: SimulationAgentDict = defaultdict(AgentData)
+            for i, state in enumerate(states_list):
+                agent_dict[str(i)] = AgentData(
+                    state=state,
+                    properties=agent_properties[i],
```
KieranRatcliffeInvertedAI: Given the addition of waypoints, this needs to be expanded to be get the agent_properties per timestep similar to agent_states.

invertedai/logs/logger.py
```diff
@@ -507,94 +646,127 @@ def initialize(
...
-                lights_random_seed=lights_random_seed,
-                initialize_random_seed=initialize_random_seed,
-                drive_random_seed=drive_random_seed,
-                initialize_model_version=init_response.api_model_version,
-                drive_model_version=drive_model_version,
-                light_recurrent_states=init_response.light_recurrent_states,
-                recurrent_states=init_response.recurrent_states,
-                waypoints_per_frame=[waypoints],
-                present_indexes=[present_indexes]
+            keys = agent_ids or [str(i) for i in range(len(init_response.agent_states))]
```
KieranRatcliffeInvertedAI: This doesn't look like proper formatting. What is the "or" supposed to be doing here? What if `agent_ids` is None?

invertedai/logs/logger.py
```diff
@@ -507,94 +646,127 @@ def initialize(
         initialize_random_seed: Optional[int] = None,
         drive_random_seed: Optional[int] = None,
         drive_model_version: Optional[str] = None,
-        scenario_log: Optional[ScenarioLog] = None,
-        waypoints: Optional[WaypointsDict] = None
-    ): 
+        scenario_log: Optional[Union[ScenarioLog, ScenarioLogV2]] = None,
+        waypoints: Optional[WaypointsDict] = None,
```
KieranRatcliffeInvertedAI, line 633: This parameter should be deprecated as unnecessary given the existence of SimulationAgentDict

invertedai/logs/logger.py
KieranRatcliffeInvertedAI: By improper formatting, I mean it is not easy to determine what is happening here at a glance. There must be a more readable, maintainable, and more robust way to do whatever this line is attempting to do.

invertedai/logs/logger.py
```diff
@@ -507,94 +646,127 @@ def initialize(
...
+            self.simulation_length = len(self._scenario_log.agent_history)
+            return
+
+        assert location is not None, "No scenario log given, must provide a location argument."
+        assert location_info_response is not None, "No scenario log given, must provide a location_info_response argument."
+        assert init_response is not None, "No scenario log given, must provide a init_response argument."
 
+        if agents_dict is not None:
+            # Use the provided keyed dictionary
+            agent_dict = agents_dict
```
KieranRatcliffeInvertedAI: What is this code block doing? It seems like it achieves nothing functionally.

invertedai/logs/logger.py
```diff
@@ -507,94 +646,127 @@ def initialize(
...
+            recurrent_states = drive_response.recurrent_states or [None] * len(current_present_indexes)
+            agent_dict: SimulationAgentDict = {}
+            for state_idx, prop_idx in enumerate(current_present_indexes):
+                key = str(prop_idx)
+                agent_dict[key] = AgentData(
+                    state=drive_response.agent_states[state_idx],
+                    properties=all_props_map[key],
+                    recurrent=recurrent_states[state_idx],
+                )
+            self._scenario_log.add_time_step_data(agent_dict)
```
KieranRatcliffeInvertedAI, line 742: This is another example of a block of code that can be replaced since current_present_indexes is superfluous now.

KieranRatcliffeInvertedAI: Insomuch as present_indexes is mostly an internal variable for tracking which agents are present at any particular timestep. This functionality is replaced by the list of SimulationAgentDicts that can be referenced fairly easily.

invertedai/logs/logger.py
```diff
@@ -507,94 +646,127 @@ def initialize(
...
+        else:
+            keys = agent_ids or list(self._scenario_log.get_agents_current().keys())
+            assert len(keys) == len(drive_response.agent_states), (
+                f"agent_ids length ({len(keys)}) must match number of agents in drive_response ({len(drive_response.agent_states)})."
+            )
```
KieranRatcliffeInvertedAI, line 753: This is not a requirement of LogWriter. In fact, this tool is designed to be agnostic to which agents are present at which timestep. Please remove all changes that were implemented with the assumption that the number of agents must be consistent for every timestep.

KieranRatcliffeInvertedAI (changes requested):

invertedai/helpers/simulation_manager.py
KieranRatcliffeInvertedAI: @atao-inverted please provide this explanation.

invertedai/logs/logger.py
```diff
@@ -79,8 +84,174 @@ def add_time_step_data(
             current_present_indexes=current_present_indexes
         )
         self.present_indexes.append(current_present_indexes)
-        self.agent_states.append(current_agent_states)        
+        self.agent_states.append(current_agent_states)
+
+    def to_scenario_log(self) -> "ScenarioLog":
```
KieranRatcliffeInvertedAI: Why is the return type here a string?

invertedai/logs/logger.py
```diff
@@ -79,8 +84,174 @@ def add_time_step_data(
...
+                        props = deepcopy(props)
+                        props.waypoints = frame_wp[str(idx)]
+                agent_dict[str(idx)] = AgentData(
+                    state=states[pos],
+                    properties=props,
+                    recurrent=None,
+                )
+            agent_data.append(agent_dict)
+
+        from invertedai.logs.logger import ScenarioLog as _ScenarioLog
```
KieranRatcliffeInvertedAI: Why is something from this exact file being imported under a different name??

invertedai/logs/logger.py
```diff
@@ -79,8 +84,174 @@ def add_time_step_data(
...
+            drive_model_version=self.drive_model_version,
+            light_recurrent_states=self.light_recurrent_states,
+            recurrent_states=self.recurrent_states,
+        )
+
+    @classmethod
+    def from_scenario_log(
+        cls, 
+        scenario_log
+    ):
```
KieranRatcliffeInvertedAI, line 212: Please provide type-hinting here. Which format of scenario_log is the input?

KieranRatcliffeInvertedAI (changes requested):

invertedai/logs/logger.py
KieranRatcliffeInvertedAI: I don't think this comment has been addressed but I think it also may not be accurate. I think if `agent_dict` is not provided (i.e. to preserve backwards compatibility), we can assume they don't care about specifying the keys of the agents. Therefore, it agent_dict is not provided AND `current_present_indexes` is not specified, assume that the agent keys have not changed from the previous time step. If the number of keys do not match, that is an error state.

KieranRatcliffeInvertedAI (changes requested):

invertedai/logs/logger.py
KieranRatcliffeInvertedAI: I don't think this comment has not been addressed.

invertedai/logs/logger.py
```diff
@@ -472,30 +618,6 @@ def export_log_to_file(
...
-                    is_waypoints = True
-                    waypoints_dict[agent_id] = prop.waypoints
-        if waypoints is not None: #Overwrite it waypoints were given explicitly
-            for agent_id, wps in waypoints.items():
-                waypoints_dict[agent_id] = wps
-
-        return waypoints_dict
     
     @validate_arguments
     def initialize(
```
KieranRatcliffeInvertedAI, line 623: Please add a note about which parameters are deprecated due to the advent of `agents_dict`. They should still be included for backwards compatibility but there should be an explicit note saying those parameters are deprecated specifically:

- waypoints
- agent_ids

invertedai/logs/logger.py
```diff
@@ -507,94 +629,130 @@ def initialize(
...
+            drive_random_seed=drive_random_seed,
+            initialize_model_version=init_response.api_model_version,
+            drive_model_version=drive_model_version,
+            light_recurrent_states=init_response.light_recurrent_states,
+            recurrent_states=init_response.recurrent_states,
+        )
+        self.simulation_length = 1
 
     @validate_arguments
     def drive(
```
KieranRatcliffeInvertedAI, line 700: Please add a note about which parameters are deprecated due to the advent of `agents_dict`. They should still be included for backwards compatibility but there should be an explicit note saying those parameters are deprecated specifically:

- waypoints
- agent_ids

KieranRatcliffeInvertedAI (changes requested):

invertedai/logs/logger.py
```diff
@@ -642,10 +642,6 @@ def initialize(
         Fallback:  pass init_response for backwards compatibility — integer string keys
                    ("0", "1", ...) will be assigned automatically.
         Legacy:    pass scenario_log (ScenarioLog) to initialize from an existing log.
-
-        Deprecated parameters (kept for backwards compatibility, use agents_dict instead):
-            waypoints: Waypoints are now stored directly in AgentProperties within the agents_dict.
-            agent_ids: Agent IDs are now the keys of agents_dict.
```
KieranRatcliffeInvertedAI: Why was the deprecation comment deleted here? Did you by chance tell Claude to revert changes and it deleted this as well?

invertedai/logs/logger.py
```diff
@@ -713,10 +709,6 @@ def drive(
         Preferred: pass agents_dict (SimulationAgentDict) with all agents for this timestep.
         Fallback:  pass drive_response with optional agent_ids and agent_properties to build the dict.
         Legacy:    pass current_present_indexes and new_agent_properties for index-based tracking.
-
-        Deprecated parameters (kept for backwards compatibility, use agents_dict instead):
-            waypoints: Waypoints are now stored directly in AgentProperties within the agents_dict.
-            agent_ids: Agent IDs are now the keys of agents_dict.
```
KieranRatcliffeInvertedAI: Why was the deprecation comment deleted here? Did you by chance tell Claude to revert changes and it deleted this as well?

KieranRatcliffeInvertedAI (changes requested):

KieranRatcliffeInvertedAI (approved):

PR #293, opened 2026-03-26
bug fixes from testing

atao-inverted:
Bug fixes from testing.

### DIFF
--- a/invertedai/helpers/simulation_manager.py
+++ b/invertedai/helpers/simulation_manager.py
@@ -11,6 +11,7 @@ from invertedai.large.initialize import large_initialize, get_regions_default, R
 from invertedai.large.drive import large_drive
 from invertedai.logs.logger import LogWriterConfig, LogWriter
 from invertedai.large.common import Region
+from invertedai.api.location import location_info
 from matplotlib.animation import FuncAnimation
 import uuid
 
@@ -39,11 +40,14 @@ class SimulationManager:
             log_writer_cfg: Optional[LogWriterConfig] = None, # can optionally initialize a log_writer_cfg to write a json file log of the simulation
         ):
             self.scene_plotter = None
+            self.scene_plotter_cfg = scene_plotter_cfg
             if scene_plotter_cfg:
+                map_center = scene_plotter_cfg.map_center if scene_plotter_cfg.map_center is not None else \
+                    (scene_plotter_cfg.location_info_response.map_center.x, scene_plotter_cfg.location_info_response.map_center.y)
                 self.scene_plotter = ScenePlotter(
                     scene_plotter_cfg.location_info_response.birdview_image.decode(),
                     scene_plotter_cfg.location_info_response.map_fov,
-                    (scene_plotter_cfg.location_info_response.map_center.x, scene_plotter_cfg.location_info_response.map_center.y),
+                    map_center,
                     scene_plotter_cfg.location_info_response.static_actors,
                     left_hand_coordinates = scene_plotter_cfg.location.split(":")[0] == "carla"
                 )
@@ -385,15 +389,27 @@ class SimulationManager:
     
     def visualize_data(self, **kwargs) -> FuncAnimation:
         """
-        Produce an animation of sequentially recorded steps. If a ScenePlotter was configured during initialization, 
+        Produce an animation of sequentially recorded steps. If a ScenePlotter was configured during initialization,
             recorded steps from each drive will be visualized using the birdview map and static actors.
-        
+
         A matplotlib animation object can be returned and/or a gif saved of the scene.
+
+        If fov or xy_offset are provided, a new birdview image will be fetched from location_info
+        to match the updated view.
 
         For kwargs, please see documentation from :func:`animate_scene` in the ScenePlotter class
         """
         if self.scene_plotter is None:
             raise ValueError("ScenePlotter not initialized, failed to animate scene")
+        if self.scene_plotter_cfg and ('fov' in kwargs or 'xy_offset' in kwargs):
+            new_fov = kwargs.get('fov', self.scene_plotter.fov)
+            new_center = kwargs.get('xy_offset', self.scene_plotter.xy_offset)
+            location_info_response = location_info(
+                location=self.scene_plotter_cfg.location,
+                rendering_fov=new_fov,
+                rendering_center=new_center,
+            )
+            self.scene_plotter.map_image = location_info_response.birdview_image.decode()
         self.scene_plotter.animate_scene(**kwargs)
     
     def export_log(self, path: Optional[str] = None):
--- a/invertedai/helpers/simulation_manager.py
+++ b/invertedai/helpers/simulation_manager.py
@@ -411,15 +411,6 @@ class SimulationManager:
         """
         if self.scene_plotter is None:
             raise ValueError("ScenePlotter not initialized, failed to animate scene")
-        if self.scene_plotter_cfg and ('fov' in kwargs or 'xy_offset' in kwargs):
-            new_fov = kwargs.get('fov', self.scene_plotter.fov)
-            new_center = kwargs.get('xy_offset', self.scene_plotter.xy_offset)
-            location_info_response = location_info(
-                location=self.scene_plotter_cfg.location,
-                rendering_fov=new_fov,
-                rendering_center=new_center,
-            )
-            self.scene_plotter.map_image = location_info_response.birdview_image.decode()
         self.scene_plotter.animate_scene(**kwargs)
     
     def export_log(self, path: Optional[str] = None):
--- a/invertedai/helpers/simulation_manager.py
+++ b/invertedai/helpers/simulation_manager.py
@@ -42,25 +42,12 @@ class SimulationManager:
             self.scene_plotter = None
             self.scene_plotter_cfg = scene_plotter_cfg
             if scene_plotter_cfg:
-                map_center = scene_plotter_cfg.map_center if scene_plotter_cfg.map_center is not None else \
-                    (scene_plotter_cfg.location_info_response.map_center.x, scene_plotter_cfg.location_info_response.map_center.y)
-                fov = scene_plotter_cfg.fov if scene_plotter_cfg.fov is not None else \
-                    scene_plotter_cfg.location_info_response.map_fov
-                birdview_image = scene_plotter_cfg.location_info_response.birdview_image.decode()
-                if scene_plotter_cfg.fov is not None or scene_plotter_cfg.map_center is not None:
-                    correct_li = location_info(
-                        location=scene_plotter_cfg.location,
-                        rendering_fov=fov,
-                        rendering_center=map_center,
-                    )
-                    birdview_image = correct_li.birdview_image.decode()
                 self.scene_plotter = ScenePlotter(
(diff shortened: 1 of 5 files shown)


invertedai/helpers/simulation_manager.py
```diff
@@ -39,11 +40,14 @@ def __init__(
             log_writer_cfg: Optional[LogWriterConfig] = None, # can optionally initialize a log_writer_cfg to write a json file log of the simulation
         ):
             self.scene_plotter = None
+            self.scene_plotter_cfg = scene_plotter_cfg
             if scene_plotter_cfg:
+                map_center = scene_plotter_cfg.map_center if scene_plotter_cfg.map_center is not None else \
+                    (scene_plotter_cfg.location_info_response.map_center.x, scene_plotter_cfg.location_info_response.map_center.y)
```
atao-inverted: Allow for custom map_center

invertedai/helpers/simulation_manager.py
```diff
@@ -424,6 +446,21 @@ def get_properties(self) -> List[AgentProperties]:
...
+    def get_property(self, agent_id: str) -> AgentProperties:
+        if agent_id not in self.agents_dict:
+            raise KeyError(f"Agent '{agent_id}' does not exist")
+        return self.agents_dict[agent_id].properties
+
+    def get_recurrent_state(self, agent_id: str) -> RecurrentState:
+        if agent_id not in self.agents_dict:
+            raise KeyError(f"Agent '{agent_id}' does not exist")
+        return self.agents_dict[agent_id].recurrent
+
```
atao-inverted, line 461: easier for insertion/deletion and identification of specific agents

invertedai/large/initialize.py
```diff
@@ -190,7 +190,7 @@ def get_number_of_agents_per_region_by_drivable_area(
     total_num_agents: Optional[int] = None,
     agent_count_dict: Optional[Dict[AgentType,int]] = None,
     random_seed: Optional[int] = None,
-    display_progress_bar: Optional[bool] = True
+    display_progress_bar: Optional[bool] = False
```
atao-inverted, line 193: bug fix for display_progress_bar

Used to always show, now defaults to not show

invertedai/large/initialize.py
```diff
@@ -431,7 +431,7 @@ def _initialize_regions(
     get_infractions: bool = False,
     random_seed: Optional[int] = None,
     api_model_version: Optional[str] = None,
-    display_progress_bar: bool = True,
+    display_progress_bar: bool = False,
```
atao-inverted, line 434: bug fix for display_progress_bar

Used to always show, now defaults to not show unless explicitly set

invertedai/large/initialize.py
```diff
@@ -561,7 +561,7 @@ def large_initialize(
...
```
atao-inverted, line 564: bug fix for display_progress_bar

Used to always show, now defaults to not show unless explicitly set

invertedai/logs/logger.py
```diff
@@ -343,13 +343,15 @@ def visualize_range(
         direction_vec: bool = False,
         velocity_vec: bool = False,
         plot_frame_number: bool = True,
-        left_hand_coordinates: bool = False,
+        left_hand_coordinates: Optional[bool] = None,
```
atao-inverted, line 346: auto-detection of 'carla' maps for left_hand coordinates

invertedai/logs/logger.py
```diff
@@ -343,13 +343,15 @@ def visualize_range(
...
         agent_ids: Optional[List[int]] = None
     ):
         """
         Use the available internal tools to visualize the a specific range of time steps within the log and save it to a given location. If
         an invalid time step range is given, the function will fail. Please refer to ScenePlotter for details on the visualization tool.
         """
+        if left_hand_coordinates is None:
+            left_hand_coordinates = self._scenario_log.location.split(":")[0] == "carla"
```
atao-inverted, line 354: auto-detection of 'carla' maps for left_hand coordinates

invertedai/logs/logger.py
```diff
@@ -413,7 +415,7 @@ def visualize(
         direction_vec: bool = False,
         velocity_vec: bool = False,
         plot_frame_number: bool = True,
-        left_hand_coordinates: bool = False,
+        left_hand_coordinates: Optional[bool] = None,
```
atao-inverted, line 418: auto-detection of 'carla' maps for left_hand coordinates

invertedai/logs/logger.py
```diff
@@ -676,6 +678,8 @@ def initialize(
                     prop = deepcopy(prop)
                     prop.waypoints = waypoints[key]
                 agent_dict[key] = AgentData(state=state, properties=prop, recurrent=rec)
+        else:
+            agent_dict = agents_dict
```
atao-inverted, line 682: This was a bug where I forgot to add the 'else' case to default to use the provided agents_dict from the parameter list

atao-inverted: This bug was evident when using SimulationManager to create the ScenarioLogs, since SimulationManager passes in the agent_dict (both external + internal agents) to LogWriter.initialize() instead of the InitilaizeResponse

invertedai/helpers/simulation_manager.py
```diff
@@ -379,15 +389,27 @@ def drive(
...
             raise ValueError("ScenePlotter not initialized, failed to animate scene")
+        if self.scene_plotter_cfg and ('fov' in kwargs or 'xy_offset' in kwargs):
+            new_fov = kwargs.get('fov', self.scene_plotter.fov)
+            new_center = kwargs.get('xy_offset', self.scene_plotter.xy_offset)
+            location_info_response = location_info(
+                location=self.scene_plotter_cfg.location,
+                rendering_fov=new_fov,
+                rendering_center=new_center,
+            )
+            self.scene_plotter.map_image = location_info_response.birdview_image.decode()
```
atao-inverted: Must call the Locationinfo again to get the correct matching birdview_image with the new requested fov and map_center

invertedai/helpers/simulation_manager.py
```diff
@@ -160,11 +163,18 @@ def _unpack(
         properties: List[AgentProperties] = []
         recurrent_states: List[RecurrentState] = []
         
+        # Determine actual recurrent size from existing agents
+        recurrent_size = RECURRENT_SIZE
+        for _, data in ordered_agents:
+            if data.recurrent is not None:
+                recurrent_size = len(data.recurrent.packed)
+                break
+
```
atao-inverted, line 170: If u add an agent mid_simulation (like during SimulationManager.drive() it has no recurrent_states, therefore we need to give it 'dummy' recurrent states that match the length of the other agents' recurrent states". Without this we would get a size mismatch error when inserting an agent mid-simulation.

invertedai/helpers/simulation_manager.py
```diff
@@ -160,11 +163,18 @@ def _unpack(
...
         for aid, data in ordered_agents:
             agent_ids.append(aid)
             states.append(data.state)  
             properties.append(data.properties)
-            recurrent_states.append(data.recurrent)  
+            recurrent_states.append(data.recurrent if data.recurrent is not None else RecurrentState(packed=[0.0] * recurrent_size))
```
atao-inverted, line 175: If u add an agent mid_simulation (like during SimulationManager.drive() it has no recurrent_states, therefore we need to give it 'dummy' recurrent states that match the length of the other agents' recurrent states". Without this we would get a size mismatch error when inserting an agent mid-simulation.

invertedai/helpers/simulation_manager.py
KieranRatcliffeInvertedAI: This seems like a duplication of code from __init__. What does this second call to location_info get that is not already provided in the call to __init__?

KieranRatcliffeInvertedAI (changes requested):
Just one change and this PR is ready to submit.

invertedai/helpers/simulation_manager.py
atao-inverted: Good question, so if the user puts custom values for `fov` and `xy_offset` we need generate the birdview with these new values, thus we must call location_info again with these new values

atao-inverted: This is because the generated birdview is static, without recalling location_info we cannot use the new `fov`/`xy_offset` (map_center) values

KieranRatcliffeInvertedAI: Ah in that case, the self.scene_plotter needs to be re-initiailized with the kwarg information as well otherwise the visualization will not be aligned. Make those changes and test the result. Also try different combinations of maps and FOV's, offsets, etc.

KieranRatcliffeInvertedAI: Or rather than re-initializing, the FOV, offset, and whatever other information needs to be modifiable. It may make sense to postpone this change until after or during the ScenePlotter Refactor. Meaning the ability to change the FOV, offset, etc. should be removed from this PR and added later.

atao-inverted: I dont think the self.scene_plotter object needs to be completely re-initilialized, but you are correct in that the 'global' values for the `fov` and `xy_offset` for self.scene_plotter must be reset. I will point out where that change was made in a separate comment

invertedai/utils.py
```diff
@@ -1085,11 +1108,23 @@ def animate_scene(
...
         """
 
+        if fov is not None:
+            self.fov = fov # reset the fov globally
+        if xy_offset is not None:
+            self.xy_offset = xy_offset
+        if fov is not None or xy_offset is not None:
+            self.extent = (- self.fov / 2 + self.xy_offset[0], self.fov / 2 + self.xy_offset[0]) + \
+                (- self.fov / 2 + self.xy_offset[1], self.fov / 2 + self.xy_offset[1])
+
```
atao-inverted: reset global values for` self.fov `and `self.xy_offset` if user chooses to override default initialization values

KieranRatcliffeInvertedAI: There is a piece of functionality that you are trying to incorporate right now which is allowing the user to input the visualization parameters at the time of visualization. There is a refactor going on currently that is changing how the ScenePlotter works. Therefore just drop this functionality for now as it's going to change immediately anyways and I don't want to temporarily commit to a supporting an architecture that is immediately about to change.

invertedai/helpers/simulation_manager.py
KieranRatcliffeInvertedAI: Based on my explanation down below, drop this functionality for now. ScenePlotter is changing too soon to commit to this.

KieranRatcliffeInvertedAI: Once ScenePlotter is refactored that allows reconfiguring visualization parameters, whether in a config parameter per visuailzation call etc., then the birdview can generated in the visualization call. Regardless of how ScenePlotter works after the refactor, the birdview should only have to be generated once. Generating it in two places would be a waste of compute when the solution is a relatively easy change of design. But again, add this later, lets not put the cart before the horse and assume what the final ScenePlotter refactor will look like at this time.

KieranRatcliffeInvertedAI (changes requested):

KieranRatcliffeInvertedAI (approved):

PR #295, opened 2026-04-07
WaypointManager EndOfRoad catch

atao-inverted:
Change the end-of-road cases to raise a silent EndOfRoadException that stops WaypointsManager from generating more waypoints for that agent

### DIFF
--- a/invertedai/__init__.py
+++ b/invertedai/__init__.py
@@ -52,7 +52,7 @@ from invertedai.logs.diagnostics import DiagnosticTool
 from invertedai.logs.debug_logger import DebugLogger
 
 # --- Helpers ---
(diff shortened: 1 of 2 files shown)


atao-inverted:
Notes for this PR:

When agents approach the end of the map, WaypointManager cannot find any more lanelets, therefore no more waypoints can be generated and currently the behaviour of WaypointManager causes this to crash with a ValueError.

There are two approaches I will try in this PR:

First approach: silently raise EndOfRoad error instead for that agent and cease waypoint generation for that agent. This causes some undefined behaviour however....

Second approach: proactively check for following lanelets and if there are none detected. If none detected raise a NoMoreLanes error and/or try and stop that agent and/or pop it from simulation

KieranRatcliffeInvertedAI:
If this PR is no longer necessary, please close it.

PR #297, opened 2026-04-10
EndOfRoadHandler tool to be used with SimulationManager

atao-inverted:
Created new class `EndOfRoadHandler` as a tool to be used within the `SimulationManager` workflow with its own config `EndOfRoadConfig`. 

This tool takes heavy inspiration from `WaypointManager` in its lanelet logic. 

At its basis it takes in `AgentStates` and a `lanelet_map` from `LocationInfoResponse` and checks for each `AgentState` whether there is a following lanelet in the `lanelet_map` given a certain spacing threshold.

### DIFF
--- a/examples/simulation_manager_minimal_example.py
+++ b/examples/simulation_manager_minimal_example.py
@@ -21,8 +21,8 @@ print("Begin initialization.")
 location_info_response = iai.location_info(location=LOCATION, include_map_source=True)
 scene_plotter_cfg = ScenePlotterConfig(location=LOCATION, location_info_response=location_info_response)
 waypoint_cfg = WaypointManagerConfig(lanelet_map = location_info_response.get_lanelet_map())
-log_cfg = LogWriterConfig(log_path="keyed_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
-simulation_manager = SimulationManager(scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg)
+log_cfg = LogWriterConfig(log_path="simulation_manager_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
+simulation_manager = SimulationManager(scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg, remove_offroad_agents=True)
 regions_config = RegionsConfig(location=LOCATION, agent_count_dict={AgentType.car: NUM_AGENTS})
 regions = simulation_manager.form_regions(regions_config)
 response = simulation_manager.initialize(location=LOCATION, regions=regions)
@@ -37,7 +37,7 @@ print("Simulation finished, save visualization.")
 
 fig, ax = plt.subplots(constrained_layout=True, figsize=(10, 10))
 simulation_manager.visualize_data(
-    output_name="Simulation_manager_minimal_example.gif",
+    output_name="simulation_manager_minimal_example.gif",
     ax=ax,
     direction_vec=False,
     velocity_vec=False,
--- a/invertedai/helpers/end_of_road_handler.py
+++ b/invertedai/helpers/end_of_road_handler.py
@@ -1,0 +1,141 @@
+from typing import Dict, List, Optional, Tuple
+from pydantic import BaseModel, ConfigDict
+from dataclasses import dataclass
+
+import lanelet2
+import numpy as np
+
+from invertedai.common import AgentState, AgentProperties, AgentData, RecurrentState, SimulationAgentDict
+from invertedai.helpers.waypoints import _find_direction_and_nearest_points
+
+_traffic_rules = lanelet2.traffic_rules.create(
+    lanelet2.traffic_rules.Locations.Germany,
+    lanelet2.traffic_rules.Participants.Vehicle
+)
+
+
+class EndOfRoadConfig(BaseModel):
+    """
+    Configuration for :class:`EndOfRoadHandler`.
+
+    Parameters:
+    lanelet_map : lanelet2.core.LaneletMapLayers
+        Projected lanelet map used to check for following lanelets.
+    waypoint_spacing : float
+        Distance threshold in meters. An agent whose best-aligned lanelet has
+        no successors and whose distance to that lanelet's endpoint is less than
+        this value is considered to be at the end of the road.
+    remove_agent : bool
+        If True, end-of-road agents are dropped from the simulation entirely 
+        they will not appear in subsequent drive calls, visualization, or logs.
+        If False, agents are frozen at their last known position and
+        continue to appear in visualization and logs.
+    """
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    lanelet_map: lanelet2.core.LaneletMapLayers
+    waypoint_spacing: float = 3.0
+    remove_agent: bool = True
+
+
+class EndOfRoadHandler:
+    """
+    Detects agents approaching the end of the road
+
+    Detection uses only the lanelet routing graph and agent states
+    
+    SimulationManager calls :func:`update` before
+    calling WaypointManager so frozen agents retain their last valid waypoints,
+    and passes the resulting IDs as an ``agents_mask`` to
+    WaypointManager to stop waypoint generation for those agents.
+    """
+
+    def __init__(self, cfg: EndOfRoadConfig):
+        self.cfg = cfg
+        self._routing_graph = lanelet2.routing.RoutingGraph(cfg.lanelet_map, _traffic_rules)
+        self._end_of_road_ids: set = set()
+        self._frozen_agents: SimulationAgentDict = {}
+
+    def is_end_of_road(self, state: AgentState) -> bool:
+        """
+        Returns True if the agent is near the end of a lanelet with
+        no following lanelets in the routing graph
+
+        inspired by 'func:generate_lane_ids_from_lanelet_map' in helpers/waypoints.py
+        """
+        x, y, yaw = state.center.x, state.center.y, state.orientation
+        filtered_lanelets = []
+        for radius in [0.0, 0.1, 0.5, 1.0, 2.0, 5.0]:
+            starting_lanelets = lanelet2.geometry.findWithin2d(
+                self.cfg.lanelet_map.laneletLayer,
+                lanelet2.core.BasicPoint2d(x, y),
+                radius
+            )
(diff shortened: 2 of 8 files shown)


invertedai/helpers/waypoints.py
```diff
@@ -128,6 +135,10 @@ def update(
             state = agent_states[i]
 
             if mask or target_paths[i] is not None:
+                if agent_properties[i].waypoints is not None and len(agent_properties[i].waypoints) == 0:
+                    props.waypoints = []
+                    _agent_properties[i] = props
+                    continue
```
KieranRatcliffeInvertedAI: This block breaks the logic of WaypointManager. An empty list indicates that the waypoints need to be refilled under normal conditions (e.g. if it reaches a single waypoint 30m ahead, it needs to be refilled and keep going).

KieranRatcliffeInvertedAI (changes requested):

invertedai/helpers/waypoints.py
```diff
@@ -517,6 +547,16 @@ def generate_lane_ids_from_lanelet_map(
                 msg=msg
             )
         return []
+    # check if the best-aligned lanelet has no successors and the agent the end of road
+    best_lanelet, _ = min(filtered_lanelets, key=lambda x: x[1])
+    if not routing_graph.following(best_lanelet, withLaneChanges=True):
```
atao-inverted: withLaneChanges should actually be false ?

examples/simulation_manager_minimal_example.py
```diff
@@ -21,8 +21,8 @@
 location_info_response = iai.location_info(location=LOCATION, include_map_source=True)
 scene_plotter_cfg = ScenePlotterConfig(location=LOCATION, location_info_response=location_info_response)
 waypoint_cfg = WaypointManagerConfig(lanelet_map = location_info_response.get_lanelet_map())
-log_cfg = LogWriterConfig(log_path="keyed_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
-simulation_manager = SimulationManager(scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg)
+log_cfg = LogWriterConfig(log_path="simulation_manager_minimal_example_log.json",location=LOCATION, location_info_response=location_info_response)
+simulation_manager = SimulationManager(scene_plotter_cfg=scene_plotter_cfg, waypoint_cfg=waypoint_cfg, log_writer_cfg=log_cfg, remove_offroad_agents=True)
```
KieranRatcliffeInvertedAI: "remove_offroad_agents" and any other flags like this should be in some sort of config.

invertedai/helpers/simulation_manager.py
```diff
@@ -49,6 +50,8 @@ def __init__(
                     scene_plotter_cfg.location_info_response.static_actors,
                     left_hand_coordinates = scene_plotter_cfg.location.split(":")[0] == "carla"
                 )
+            self.remove_offroad_agents = remove_offroad_agents
+            self._offroad_agent_ids: set = set()
```
KieranRatcliffeInvertedAI: A set is an okay structure to keep track of agents but I think it should be wrapped into a more general structure that tracks any special simulation behaviour (e.g. end of road behaviour or future additions like a weather parameter as an example - not that that is something we are adding at this time).

KieranRatcliffeInvertedAI (changes requested):
I think any wording for "offroad" should be changed to "end of road" because offroad is a bit too general.

invertedai/simulation_manager.py
```diff
@@ -357,13 +365,20 @@ def drive(
...
             properties = self.waypoint_manager.update(
-                response = response,
-                agent_properties = properties,
+                response=response,
+                agent_properties=properties,
+                agents_mask=agents_mask,
             )
+            #Detect off-road agents (EndOfMapException sets waypoints=[])
+            for i, aid in enumerate(agent_ids):
+                if aid not in external_ids and properties[i].waypoints is not None and len(properties[i].waypoints) == 0:
```
KieranRatcliffeInvertedAI: This condition will result in many false positives. Agents reaching the end of their assigned waypoints happens very often and does not indicate coming to the end of the road.

invertedai/helpers/waypoints.py
```diff
@@ -138,7 +145,9 @@ def update(
                             target_path = target_paths[i],
                             agent_properties = props
                         )
-                    
+                        if len(props.waypoints) == 0:
```
KieranRatcliffeInvertedAI: This condition is not true. Agents reach the end of their waypoint list all the time. It does not indicate that they've reached the end of their waypoint list. E.g. it could simply mean it's achieved its given set of waypoints in the middle of a closed, Carla map and needs to be given a new list.

invertedai/helpers/waypoints.py
```diff
@@ -153,6 +162,8 @@ def update(
                         #Do not pass the original target path as it should be completed if the list is empty
                         waypoint_flags.append(WaypointUpdateFlags.WAYPOINTS_EMPTY)
                         props.waypoints = self.generate_waypoints(state=state)
+                        if len(props.waypoints) == 0:
```
KieranRatcliffeInvertedAI: This condition is not true. Agents reach the end of their waypoint list all the time. It does not indicate that they've reached the end of their waypoint list. E.g. it could simply mean it's achieved its given set of waypoints in the middle of a closed, Carla map and needs to be given a new list.

invertedai/helpers/waypoints.py
```diff
@@ -165,6 +176,13 @@ def update(
                             target_path = target_paths[i],
                             agent_properties = props
                         )
+                        if len(props.waypoints) == 0:
```
KieranRatcliffeInvertedAI: This condition is not true. Agents reach the end of their waypoint list all the time. It does not indicate that they've reached the end of their waypoint list. E.g. it could simply mean it's achieved its given set of waypoints in the middle of a closed, Carla map and needs to be given a new list.

invertedai/helpers/waypoints.py
```diff
@@ -565,6 +605,10 @@ def generate_lane_ids_from_lanelet_map(
             ending_lanelets = sorted(list(routing_graph.reachableSet(starting_lanelet, maxRoutingCost=maxRoutingCost, allowLaneChanges=True)), key=lambda lanelet: lanelet.id)
             total_distance += route.length2d()
             route_lane_ids.extend([lanelet.id for lanelet in list(route.shortestPath())[1:]])
+        if not ending_lanelets and total_distance < (waypoint_spacing if waypoint_spacing is not None else 20.0):
```
KieranRatcliffeInvertedAI: Why would waypoint_spacing ever be None? Why make this parameter Optional?

invertedai/helpers/waypoints.py
```diff
@@ -517,6 +547,16 @@ def generate_lane_ids_from_lanelet_map(
                 msg=msg
             )
         return []
+    # check if the best-aligned lanelet has no successors and the agent the end of road
+    best_lanelet, _ = min(filtered_lanelets, key=lambda x: x[1])
+    if not routing_graph.following(best_lanelet, withLaneChanges=True):
+        end_point = best_lanelet.centerline[-1]
+        dist_to_end = np.sqrt((x - end_point.x)**2 + (y - end_point.y)**2)
+        if dist_to_end < (waypoint_spacing if waypoint_spacing is not None else 30.0):
```
KieranRatcliffeInvertedAI: Why would waypoint_spacing ever be None? Why make this parameter Optional?

invertedai/helpers/waypoints.py
```diff
@@ -469,19 +497,21 @@ def generate_lane_ids_from_lanelet_map(
     min_distance: Optional[float] = 1000.0, 
     destination_waypoint: Optional[Point] = None,
     seed: Optional[int] = None,
-    logger: Optional[logging.Logger] = None
+    logger: Optional[logging.Logger] = None,
+    waypoint_spacing: Optional[float] = 20.0
```
KieranRatcliffeInvertedAI: The default value of waypoint_spacing needs to be 30.0m for performance reasons.

invertedai/helpers/waypoints.py
KieranRatcliffeInvertedAI: I can see how both cases can cause complications. I suppose it depends on the end-of-road behaviour for whether it should check for lange changes or not. For instance, if the dist_to_end threshold to be considered at the end of the road is relatively large (which I would say 30m is) withLaneChanges should be True because it could theoretically change into a lane that isn't ending in that space. But if the dist_to_end value is low (lets say 15m), then we could set withLaneChanges to False and that will probably reduce latency with similar performance. WaypointManager is already fairly bulky as is so the latter would be my preference.

invertedai/helpers/end_of_road_handler.py
```diff
@@ -0,0 +1,137 @@
...
+            for _, ll in sorted(starting_lanelets, key=lambda l: l[1].id):
+                a, b = _find_direction_and_nearest_points(
+                    ll.centerline,
+                    lanelet2.core.BasicPoint3d(x, y, 0)
+                )
+                lane_orientation = np.arctan2(b.y - a.y, b.x - a.x)
+                angle = np.absolute((yaw - lane_orientation + np.pi) % (2 * np.pi) - np.pi)
+                if angle < 75 * np.pi / 180:
+                    filtered_lanelets.append((ll, angle))
+            if filtered_lanelets:
```
atao-inverted: Only Code it Once waypoints.py

invertedai/helpers/waypoints.py
```diff
@@ -610,6 +591,30 @@ def _find_direction_and_nearest_points(
...
+        )
+        for _, lanelet in sorted(starting_lanelets, key=lambda l: l[1].id):
+            a, b = _find_direction_and_nearest_points(lanelet.centerline, lanelet2.core.BasicPoint3d(x, y, 0))
+            lane_orientation = np.arctan2(b.y - a.y, b.x - a.x)
+            angle = np.absolute((yaw - lane_orientation + np.pi) % (2 * np.pi) - np.pi)
+            if angle < 75 * np.pi / 180:
+                filtered_lanelets.append((lanelet, angle))
+        if filtered_lanelets:
+            break
+    return filtered_lanelets
```
KieranRatcliffeInvertedAI: I'd say keep the error msg functionality for this message and pass it as one of the output parameters. When this function is applied to invertedai/helpers/waypoints.py, just make the msg output an underscore "_" to imply that it is not used.

KieranRatcliffeInvertedAI (approved):

